PR #14, opened 2017-01-11
Thrussh is completely based on tokio since 0.8

P-E-Meunier:
Thrussh is an SSH client and server library. The API has been tested with other protocols like postgres, using asynchronous callbacks spawn in the same event loop.

Tokio is awesome! Thanks a lot!


aturon:
Wonderful, thanks for the PR!

alexcrichton:
Awesome, thanks @P-E-Meunier! Want to write a bit of a blurb in the bulleted list above as well? (I think right now it won't show up as it's just a markdown link)

PR #15, opened 2017-01-11
third-party: Fix a typo and reword inotify/signal

cgwalters:
"allows abstraction Unix signals" doesn't parse, and I think it's simpler to
make both the inotify and signal text consistent and shorter ("maps" rather than
"enables use of").


aturon:
Thanks!

PR #16, opened 2017-01-11
Indicate current page in single page/blog post side nav

acobster:
All links in the sidenav currently get the `text-muted` class. This threw me off a bit when I was browsing through documentation.


aturon:
Lovely, thank you!

PR #24, opened 2017-01-11
Add link for blog RSS feed so browsers can see it

### LINKED ISSUE #10: Blog should have Atom or RSS feed
I just discovered the new Tokio website today and saw that there's an associated blog. I'd like to keep up with developments in Tokio, but it seems there's no Atom or RSS feed for the blog that would make that easy to do.

I don't know what that would take in Hugo to implement, so this is more of a "nice to have" than anything.

schuster:
Fixes #10 . To see the change, open the website in Firefox and note that the "Subscribe to This Page..." option is enabled under the Bookmarks menu.

This puts the link on all pages by default. I don't see any harm in that, but if it should only be on the blog pages, that probably requires some conditional formatting in Hugo that I don't know how to do yet.

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -12,6 +12,9 @@
     <link rel="stylesheet" href="/css/font-awesome.min.css">
     <link rel="stylesheet" href="/css/tokio.css">
 
+    <!-- Blog feed link -->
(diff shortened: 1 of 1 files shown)


alexcrichton:
Sounds good to me, thanks!

PR #30, opened 2017-01-12
Various sundry typos and whatnot

rbalicki2:
Various things I came across while reading the docs


aturon:
Wow, thanks much!

rbalicki2:
ah rats I had one follow up that I got after it was merged! Oh well!
I learned a lot from reading these docs, thanks! :+1:

PR #31, opened 2017-01-12
fix(styles): changes to accommodate mobile

jefflembeck:
This includes changes to the header, the main section of the homepage,
and the jumbotron section in order to better accommodate mobile browsers

Addresses #11 in a few ways

### DIFF
--- a/static/css/tokio.css
+++ b/static/css/tokio.css
@@ -11,7 +11,11 @@ table, td {
 }
 
 td {
-    padding: 4px
+    padding: 4px;
+}
+
+img {
+  max-width: 100%;
 }
 
 /*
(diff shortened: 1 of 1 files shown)


ashleygwilliams:
these changes look awesome. for those curious, among other things, it makes the mobile view go from 
screen shot 2017-01-12 at 7 26 57 am
screen shot 2017-01-12 at 7 26 16 am

ashleygwilliams:
@jefflembeck maybe for the mobile view, remove the logo from the nav list? just a thought

carllerche:
Definitely a solid improvement. Thanks. I have merge: 8c78712901d3a430ebe0ec8e8147d5b92f3dd291

I do think that we could probably do some additional tweaks to the logo. Also, it would be nice if the menu and TOC on the doc page could be collapsed on mobile :)

PR #41, opened 2017-01-15
More explicit explanation about unparks

vorner:
I tried adding some a bit more explicit notes about how the executors work, explaining there's less magic than I previously though. I believe these would have saved me several hours of trying to read the code, running examples through gdb and asking.

But I understand this is very personal, so this doesn't necessarily have to be better in your opinion.

### DIFF
--- a/content/docs/getting-started/futures.md
+++ b/content/docs/getting-started/futures.md
@@ -205,6 +205,8 @@ use std::time::Duration;
 
 use futures::Future;
 use futures_cpupool::CpuPool;
+// There's also a tokio_core::reactor::Timeout, but it needs an event loop
+// to run on.
 use tokio_timer::Timer;
 
 # const BIG_PRIME: u64 = 1;
--- a/content/docs/getting-started/futures.md
+++ b/content/docs/getting-started/futures.md
@@ -205,8 +205,6 @@ use std::time::Duration;
 
 use futures::Future;
 use futures_cpupool::CpuPool;
-// There's also a tokio_core::reactor::Timeout, but it needs an event loop
-// to run on.
 use tokio_timer::Timer;
 
 # const BIG_PRIME: u64 = 1;
--- a/content/docs/going-deeper/tasks.md
+++ b/content/docs/going-deeper/tasks.md
@@ -151,7 +151,14 @@ call park, if an unpark happens it'll store `true` to `ready` (as we saw above
 in `Unpark for ThreadUnpark`). If this hasn't happened, though, then we block
 the current thread. The current thread will then get reawoken once the future's
 task is unparked, through the `ThreadUnpark` instance associated with this
-future.
+future. This must be done from another thread that is not parked. Some futures
+need a `Core` to be running in some thread to proceed, or some other trigger
+they registered into. Obviously a different one than which we block with `wait`.
+
+If you remember the
+[example with prime verification](../getting-started/futures#adding-a-timeout),
+it can work only because both the computation and the timeout run in separate
+threads, waking up the waiting future.
 
 [`poll_future`]: https://docs.rs/futures/0.1/futures/executor/struct.Spawn.html#method.poll_future
 [`poll`]: https://docs.rs/futures/0.1/futures/future/trait.Future.html#tymethod.poll
@@ -325,7 +332,7 @@ leveraged the new [`execute`] method to run the future instead. This method is
 more restrictive on the types of futures that it can accept, but it works by
 essentially submitting work to an executor whenever it's available. Futures that
 block will be resubmitted to the queue once they've been unblocked through the
-`unpark` method.
+`unpark` method, for example by an event loop in another thread.
 
 ## [Executors in `tokio-core`](#tokio-core) {#tokio-core}
 


content/docs/getting-started/futures.md
```diff
@@ -205,6 +205,8 @@ use std::time::Duration;
 
 use futures::Future;
 use futures_cpupool::CpuPool;
+// There's also a tokio_core::reactor::Timeout, but it needs an event loop
+// to run on.
```
alexcrichton: Perhaps this could be delayed to later or explained elsewhere? I find that comments like this unfortunately detract from the example at hand much of the time.

content/docs/going-deeper/tasks.md
```diff
@@ -151,7 +151,14 @@ call park, if an unpark happens it'll store `true` to `ready` (as we saw above
 in `Unpark for ThreadUnpark`). If this hasn't happened, though, then we block
 the current thread. The current thread will then get reawoken once the future's
 task is unparked, through the `ThreadUnpark` instance associated with this
-future.
+future. This must be done from another thread that is not parked. Some futures
+need a `Core` to be running in some thread to proceed, or some other trigger
+they registered into. Obviously a different one than which we block with `wait`.
```
alexcrichton: This isn't quite true I think because it's definitely possible to have a single-threaded server (just one `Core`) and tasks are all unparked there?

vorner: This comment is only about the `wait` executor, which calls `thread::park` internally. Unless I'm very mistaken, if you wait on a future, you can't run a `Core` at the same thread at that time.

Or is there some more magic I don't see?

content/docs/getting-started/futures.md
vorner: OK, how about the fixup commit? (I'd squash it once we agree if that's OK)

I'm a bit worried people would start using the `tokio-timer` crate for their general timeouts as they see in the example and I'm afraid starting another thread is an overkill most of the time, so I want to point the existence of the core timeouts out.

content/docs/getting-started/futures.md
```diff
@@ -194,7 +194,12 @@ printing a message, here), and retrieve the value later on.
 So far this example isn't terribly interesting, since there are simpler ways to
 work with thread pools. But one strength of futures is their ability to
 *combine*. We'll show this off by combining the thread pool future with a
-timeout future:
+timeout future.
+
+We'll use a timer from the `tokio-timer` crate, which manages its timeouts in a
+separate thread. There are also timeouts in `tokio-core`. They don't need
+another thread, but they need an event loop, something we'll have a look at
+later.
```
alexcrichton: This seems like it may not be that relevant to explaining of the next example? This is one of the first documents seen by those interacting with features, so we just want to make sure it's as straightforward and as clear as possible.

content/docs/going-deeper/tasks.md
alexcrichton: > This must be done from another thread that is not parked

This isn't a correct statement as threads aren't parked, *tasks* are parked. You can unpark a task from any thread, including the thread it may originally have been pinned on.

content/docs/getting-started/futures.md
vorner: OK, let's drop this one altogether. It's probably not that important anyway.

content/docs/going-deeper/tasks.md
vorner: Yes, tasks are parked. However, if I follow the code correctly, in the case of the `wait` executor in this chapter, parking the task means parking the underlying thread. Citing the paragraph just above this modified one:

| Recall that we're just blocking the current thread until the future is
| finished, so we're just going to call [`thread::park`] here from the standard
| library. To get a better idea, let's take a look at the `ThreadUnpark::park`
| function we're calling:

And the code snippet also speaks quite clearly (also just from above of the paragraph I modify):

```rust,ignore
fn park(&self) {
    if !self.ready.swap(false, Ordering::SeqCst) {
        thread::park();
    }
}
```

Indeed, in the case of different executor (the `Core`, for example), parking the task doesn't mean parking the thread and the task can be unparked from the same thread. But this modification is about the [`wait` executor](https://tokio.rs/docs/going-deeper/tasks/#exploring-wait).

The problem I'm trying to solve here is, when I first read this part of docs and remembered the example with prime computation and timeout, I wondered how on Earth can this task ever continue, when I park the underlying thread ‒ and the answer to that is its `unpark` got called from another thread (either one with the computation, or one with the timeout). My confusion probably stemmed from the idea that I'm doing all this futures dancing because I want to *avoid* having other threads around, so why would I want to park my (only) thread ‒ but that kind of thinking is relevant to the `Core` executor, not here.

alexcrichton:
@vorner I think I may be confusing myself unfortunately. Can you clarify what this is clarifying? I'm hesitant to throw too many concepts into the getting started docs as these want to be as lightweight as possible

vorner:
Well, the outstanding changes are in the going deeper (I removed the one from the getting started already).

And for what it is clarifying… When I first looked into futures and tokio, I found them very magical. I don't mean the idea of a result being available sometime in the future, but how they actually work. It's a lot of details. Basically, if you have a callback-based interface, it is obvious how it works. With these futures, it is less so ‒ all the feeling that they are actually upside down (being polled from the end result, not propagated in a chronological order), I was expecting magic on all levels.

After working through the docs and parts of the code, the only question left was what magic wakes my only thread if I parked it, if I used the `wait` executor. The way of unparking the futures in reactor is easy to understand.

It's really not obvious that `tokio-timer` starts a new thread and that there's actually another thread to wake this one, so there's no real magic ‒ `wait()` simply doesn't work properly in a single-threaded application. But I had to compile the mentioned example and step through it in gdb to find that out. I expanded the text in a way that I think would have had helped me at that time.

However, it is possible only I have this problem, or that my explanation doesn't really help and makes it even worse to others ‒ writing a good documentation is a hard task. I have no problem dropping this if it is the case.

alexcrichton:
Ok thanks for the clarification! I wonder if most of the confusion stemmed from trying to understand how tokio-timer worked? Maybe we could assist that with a few words when tokio-timer was introduced? (I think that could still fit well even in the getting started section)

vorner:
Well, not exactly, I never even suspected tokio-timer to start a new thread, so I didn't try to understand that at the time. I suspected magic in the `wait` executor and couldn't find it.

But I guess if I knew upfront that tokio-timer starts the thread, I wouldn't try so hard blaming the `wait` and would be less confused.

That being said, overall the documentation is IMO very good.

Do you think you could add that few words? Obviously, I'm not very good at formulating them in a non-intrusive way.

alexcrichton:
Ok yeah sure I can try to add some to where we introduce tokio-timer, just want to confirm that'd work well for your original intention here.

vorner:
Cool. Will you close this PR once that is done? Or let me know, so I can close it?

alexcrichton:
@vorner how does https://github.com/tokio-rs/website/pull/50 look to you?

vorner:
Thanks, that one is probably good enough. I'm going to close this one, then.

PR #48, opened 2017-01-21
Minor code cleanup in LineCodec example

mbrubeck:
* Remove an unncessary `return` keyword
* Remove an unnecessary `&` wrapper
* Use `EasyBuf::as_slice` everywhere (instead of `as_slice` once and `as_ref` once)


carllerche:
Thanks, merged 2e89eb27289a561a63610b148b7b248bfc0ad591

PR #49, opened 2017-01-21
Use `extend` instead of `extend_from_slice`

mbrubeck:
These are identical since rust-lang/rust#37094 landed, and
`extend_from_slice` will be deprecated in the future.


carllerche:
👍 would you be able to check [tokio-line](https://github.com/tokio-rs/tokio-line) to see if it uses `extend_from_slice` as well?

PR #53, opened 2017-02-20
variable does not need to be mutable.

AaronStGeorge:
During compilation a friendly warning informed me that the `service` variable does not need to be mutable.

```
warning: variable does not need to be mutable, #[warn(unused_mut)] on by default
  --> src/main.rs:58:13
   |
58 |         let mut service = s.new_service()?;
   |             ^^^^^^^^^^^
```


carllerche:
Thanks for the PR!

PR #55, opened 2017-02-25
Clarify where to replace Echo with EchoRev

danobi:
It was a little ambiguous where to do the replacement before.


alexcrichton:
Thanks!

PR #58, opened 2017-03-06
add tokio-retry to third-party crates

srijs:
I just released the second version of this crate, which is still in an earlier stage of development. Please let me know if there's a level of maturity you require of the crates listed here, and I can re-file the PR at a later stage when the crate has achieved greater stability.


alexcrichton:
Thanks!

carllerche:
A quick note re: the name. Given that cargo doesn't providing namespaces, we've been using tokio-* as an advice namespace. At some point in the future, we may want to provide an official tokio-retry crate. At which point we'll ask to take over the crate name.

Others have opted to use a different naming pattern to avoid this potential naming conflict (tk-name or name-tokio)

srijs:
@carllerche Totally understood about the name. I'd even go one step further and say that if you would like to provide an official `tokio-retry` crate, I would love to work with you to move the existing `tokio-retry` into your org and give you ownership over it. Is that something worth considering?

PR #70, opened 2017-04-07
Add links to reference documentation

### LINKED ISSUE #69: Reference Documentation
I think it would be helpful to link directly to the reference documentation for the official Tokio crates. Right now you have to do a search on docs.rs or crates.io and potentially filter through many third-party crates to find what you're looking for.

For example, look at the "Documentation" page https://tokio.rs/docs/getting-started/tokio/ . Somewhere in each of the `tokio-proto`, `tokio-core` and `futures` list items there could be links to https://docs.rs/tokio-proto, https://docs.rs/tokio-core, and https://docs.rs/futures respectively.

Another option might be to add links on the left side of the page under a new "Reference" heading.

I'm willing to work on this if there is any interest.

jbendig:
Fixes #69

### DIFF
--- a/content/docs/getting-started/tokio.md
+++ b/content/docs/getting-started/tokio.md
@@ -8,22 +8,26 @@ weight = 0
 
 It's broken into several layers, with entry points based on your needs:
 
-* **The `tokio-proto` layer is the easiest way to build servers and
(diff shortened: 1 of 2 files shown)


carllerche:
Thanks!

PR #72, opened 2017-04-20
Split "Going Deeper" section to make it easier to study

munckymagik:
# What?

Reorganises the "Going Deeper" section into two:

* "Going Deeper: Futures"
* "Going Deeper: Tokio"

# Why?

When I found the docs on tokio.rs I was really happy. It was genuinely fun working through the Getting Started section. But I'm finding it hard getting beyond this. One of the problems I noticed is the way the Futures and Tokio content is interleaved has made it a little difficult to absorb and build strong conceptual models.

Tokio is really fascinating but it's a bit ahead of where I am right now. What I want to do is learn the basic facilities for working with async code in Rust first. I've experimented with threads and channels, now I want to study Futures.

So I'm suggesting we separate the two topics. I'm also wondering whether to split the Tokio section into "Going Deeper: tokio-proto" and ""Going Deeper: tokio-core", which would continue the model introduced in "What is Tokio?".

# What does it look like?

screencapture-localhost-1313-docs-going-deeper-futures-futures-mechanics-1492719361214

### DIFF
--- a/content/docs/getting-started/core.md
+++ b/content/docs/getting-started/core.md
@@ -2,7 +2,7 @@
 title = "High-level I/O using core"
 description = "Future, stream and sink-based APIs from tokio-io and tokio-core"
 menu = "getting_started"
-weight = 6
+weight = 106
 +++
 
 At this point, we've got a pretty solid grasp of what event loops are and how
--- a/content/docs/getting-started/db.md
+++ b/content/docs/getting-started/db.md
@@ -2,7 +2,7 @@
 title = "Example: serving database content using proto"
 description = "Writing a `fortune` server using proto"
 menu = "getting_started"
-weight = 3
+weight = 103
 +++
 
 In our previous example, an echo server, we didn't leverage the fact that a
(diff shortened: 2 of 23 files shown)


alexcrichton:
Sounds like a good idea to me!

Thoughts @carllerche?

carllerche:
Yeah, that looks good to me.

We've been thinking about how to restructure the docs... but we're mostly trying to figure out how to reorganize "getting started". I think a change like this would be a good incremental step and wouldn't conflict too much w/ anything else we might do in the short term.

munckymagik:
@alexcrichton @carllerche Sorry for the delay in replying. Great news!

munckymagik:
Sorry, I hadn't realised the embedded docs in futures-rs reference this site. I just discovered this now. I assume it'll be the same for several of the tokio-* projects too?

I'm going to experiment with [Hugo aliases](https://gohugo.io/extras/aliases/) to see if we can continue to support the old URLs. I'll push a commit if that work out.

munckymagik:
Ok, Hugo aliases worked out fine.

alexcrichton:
Oh whoa awesome! I'd also be fine just updating the links elsewhere, but aliases are also great!

PR #73, opened 2017-04-23
Remove the extra `BufMut` import.

hockeybuggy:
In the simple server example `BufMut` is imported from the bytes crate, but isn't used (`BytesMut` is the only trait needed). Removing the import might improve clarity, but there may have been another reason for the import that I am not aware of.

Thanks. 😄


hockeybuggy:
Thanks! 😄

PR #74, opened 2017-04-23
Clarify the comments above the type aliases in the `ServerProto` impl block

hockeybuggy:
In the simple server example there are two comments that where not clear to me. This PR updates the comments to be more in line with my understanding (which may or may not be correct). If this is not quite right I would be happy to update it, or just have this closed.

Thanks 😄


alexcrichton:
Looks great!

PR #75, opened 2017-05-08
Switch from `rustc_seralize` to Serde.

hockeybuggy:
The `rustc_serialized` crate was deprecated in favour Serde. This PR switches the [`/getting-started/db/` example](https://tokio.rs/docs/getting-started/db/) to use Serde.

The tutorial [before](https://github.com/tokio-rs/website/blob/bf22c0cf57c230805cd5ca52a510dbf1c2c308d7/content/docs/getting-started/db.md) and [after](https://github.com/hockeybuggy/website/blob/7a5d518749d66910c8f9e0ea84958d681e04b5e3/content/docs/getting-started/db.md).


content/docs/getting-started/db.md
```diff
@@ -282,7 +291,7 @@ struct Server {
     db_pool: r2d2::Pool<r2d2_postgres::PostgresConnectionManager>,
 }
 
-#[derive(RustcEncodable)]
+#[derive(Serialize, Deserialize)]
```
alexcrichton: I think we just need `Serialize` here, right?

content/docs/getting-started/db.md
```diff
@@ -61,10 +61,15 @@ tokio-minihttp = { git = "https://github.com/tokio-rs/tokio-minihttp" }
 r2d2 = "0.7"
 r2d2_postgres = "0.11"
 
-# misc support for thread pools, random numbers, and json
+# json
+#[macro_use]
```
alexcrichton: I think this annotation can be omitted?

hockeybuggy: I get this error with the `#[macro_use]` removed: 

```
warning: proc macro crates and `#[no_link]` crates have no effect without `#[macro_use]`
 --> src/main.rs:9:1
  |
9 | extern crate serde_derive;
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^

error: cannot find derive macro `Serialize` in this scope
  --> src/main.rs:31:10
   |
31 | #[derive(Serialize)]
   |          ^^^^^^^^^
```

alexcrichton: Oh sure yeah but that's in the code, right? This one's in the TOML

hockeybuggy: :man_facepalming:

hockeybuggy: Good catch. 😄

PR #81, opened 2017-05-16
Cheatsheet for Futures

Rufflewind:
[Per suggestion](https://www.reddit.com/r/rust/comments/6bbofo/cheatsheet_for_future_combinators/dhm6sy3/) by @alexcrichton, this is a PR to integrate “Cheatsheet for Futures” into the website.  The Python script is used to add HTML formatting to the cheatsheet.

I’ve not yet actually done that in this PR, since I’m not sure which section this belongs to or even how this should be integrated.  The cheatsheet would look best if it’s rendered at maximum width with no wrapping – it’s almost like an image/diagram.  One potential idea is to have an “embedded” version (shrunk or boxed inside a scrollable div) as part of the main site, with a link pointing to the full standalone version.

Any ideas or preferences?


alexcrichton:
Thanks! I wonder if we could link to this as part of "Essential Combinators"? Either at the beginning or the end to browse through it.

If it looks fine rendered inline I'd be ok with that, but I'd also be fine if we just link to a separate page dedicated for this.

Rufflewind:
OK, I’ve added a link in the first paragraph of “Essential Combinators”:

> … It's also worth spending some time with the trait documentation to familiarize yourself with the full range of combinators vailable ([cheatsheet](/img/diagrams/cheatsheet-for-futures.html)).

alexcrichton:
Oops sorry it took awhile, but looks great to me, thanks again @Rufflewind!

PR #82, opened 2017-05-24
Populate database in getting started db example.

hockeybuggy:
This PR attempts to improve on the database example in the getting started section by:

1. Adding a section of creating the table that the example will query from
1. Switching from using a `randomNumber` column in a `World` table to a `body` column from a `greetings` table.

The example will now return a random greeting from the table and return a response such as:

```json
{"id":3,"body":"Bonjour"}
```

### DIFF
--- a/content/docs/getting-started/db.md
+++ b/content/docs/getting-started/db.md
@@ -177,7 +177,8 @@ flexibly define the future internally.
 [`BoxFuture`]: https://docs.rs/futures/0.1/futures/future/type.BoxFuture.html
 
 
-Well also need a `Message` struct that we can use to serialize the row that we will be getting from the database into json using `serde`:
+Well also need a `Message` struct that we can use to serialize the row that we
+will be getting from the database into json using `serde`:
 
 ```rust,ignore
 #[derive(Serialize)]
@@ -186,6 +187,7 @@ struct Message {
     body: String,
 }
 ```
+
 
 Next up, let's start implementing `call`. First we can verify that the HTTP
 request is indeed `/db`:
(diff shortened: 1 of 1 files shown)


content/docs/getting-started/db.md
```diff
@@ -176,6 +176,19 @@ flexibly define the future internally.
 
 [`BoxFuture`]: https://docs.rs/futures/0.1/futures/future/type.BoxFuture.html
 
+
+Well also need a `Message` struct that we can use to serialize the row that we
+will be getting from the database into json using `serde`:
```
hockeybuggy: Is it worthwhile to link to serde here?

content/docs/getting-started/db.md
```diff
@@ -188,7 +201,7 @@ just panic if you request a different path. Next up let's generate the id of
 the row that we're going to look up:
 
 ```rust,ignore
-let random_id = rand::thread_rng().gen_range(0, 10_000);
+let random_id = rand::thread_rng().gen_range(1, 5);
```
hockeybuggy: In Postgres the `serial` type starts at 1 and the example has 5 rows in the `greetings` table.

I briefly considered adding a query when the program starts, to find the number of rows and use it as an upper bound of the random number, but I concluded that it might make the example more complex than necessary.

content/docs/getting-started/db.md
```diff
@@ -285,15 +318,15 @@ use tokio_minihttp::{Request, Response};
...
-}
-
 #[derive(Serialize)]
 struct Message {
     id: i32,
-    randomNumber: i32,
+    body: String,
+}
+
+struct Server {
```
hockeybuggy: I switched the order of the `Message` struct and the `Server` struct so that the `Server` would be defined right before it's `impl`.

alexcrichton:
Ah sorry for the delay this looks great to me, thanks @hockeybuggy!

content/docs/getting-started/db.md
alexcrichton: I'd be fine either way, but if you'd like to send a follow-up PR please feel free!

hockeybuggy:
Thank you @alexcrichton 😄

PR #83, opened 2017-05-29
Added tokio-core dependency

schleumer:
While reading the docs and examples i came with an error:
```
  |
3 | extern crate tokio_core;
  | ^^^^^^^^^^^^^^^^^^^^^^^^ can't find crate
```

Then i noticed that `tokio-core` wasn't referenced on the doc's dependencies.


alexcrichton:
Thanks!

PR #87, opened 2017-06-19
The requestId has already been red from the buffer

baloo:
The buffer does not include the request id anymore, we should read the full buffer. The current implementation removes the first 4 bytes from the line.


alexcrichton:
Thanks!

PR #89, opened 2017-07-03
Guidance on where to put state and configuration

zmanian:
I found where to put the stateful bits of a server a bit confusing going through the Tokio docs that first time. i think a hint like this would have set me in a better direction


alexcrichton:
Thanks!

PR #101, opened 2017-08-27
Fix the non-code portion of the Sender::complete removal.

nelsonjchen:
Continuing commit 6a4cb84035317af177066f557911ca68c7762523

- - -

On that note, some static site generators check if linked pages in content are dead or not. But I don't know if any check if an _anchor_ exists or not nor if Hugo has this feature and it is enabled.


alexcrichton:
Thanks!

PR #104, opened 2017-08-27
Remove deprecated BoxFuture and boxed from DB example

nelsonjchen:
Reference:

https://github.com/alexcrichton/futures-rs/pull/556


alexcrichton:
Awesome thanks!

PR #105, opened 2017-08-28
Add a lot of "# #![deny(deprecated)]" and update the deprecations

nelsonjchen:
Whatever was run by `ci/test.sh`, I added `# #![deny(deprecated)]` and fixed up any deprecation errors and whatever prose I could find.

Some documented code wasn't run and specced to `ignore`, but whatever usage of `BoxFuture` I saw in those, I removed. Maybe for these `ignore`d blocks, whatever we can, we should try and get going under tests.

This was mostly removing `BoxFuture` stuff with one instance of `# #![deny(deprecated)]` catching a deprecation and refactoring that split the `IO` trait. 

What a handy lint configuration. PITA to copy and paste it everywhere for it to be used though.


alexcrichton:
Awesome, thanks so much!

PR #115, opened 2017-10-01
Changed the wordings a bit

toddWannaCode:
I think this helps in conveying the message a bit more clearly.


alexcrichton:
Thanks!

PR #116, opened 2017-10-23
[Minor] Replace doc comments with regular ones

lazyval:
Most of the documentation uses regular comments except for several occurrences of doc comments and seems like there is no reason for later (because documentation is not going to get it's own documentation).


alexcrichton:
Thanks!

PR #120, opened 2017-11-29
Import required trait into example

kanerogers:
This one tripped me a up a little bit. 😄


alexcrichton:
Thanks! I think this is intended to be a snippet and/or standalone example though? Not necessarily entirely self contained?

kanerogers:
You might be right! I was just following along at home, pasted it into my
editor and was surprised when the compiler got angry with me. Just my
feedback. 😊

carllerche:
This guide has been removed, so I will close this. Thanks!

kanerogers:
Awesome. PS. thanks for all your hard work on Tokio @alexcrichton and @carllerche - can't wait to see what comes out of the reform process. 😄

PR #121, opened 2017-12-09
Fix links to Stream::and_then and Sink::send_all methods in "Pipeline server"

funkill:
Change method `Stream::send_all` to `Sink::send_all` and add link
Fix link for `Stream::and_then` method


alexcrichton:
Thanks!

PR #142, opened 2018-02-20
Fix a bug in Blog menu order

Yoopo:
Calling reverse on Site Menu with out sorting it lead to strange behaviour.
The menu order is not the same on every blog page.
The last post link to the first one when visiting last post page.

### DIFF
--- a/layouts/blog/single.html
+++ b/layouts/blog/single.html
@@ -15,7 +15,7 @@
       <nav class="leftnav">
         <div class="" id="">
           <ul class="nav">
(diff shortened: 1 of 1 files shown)


carllerche:
Hmm.. in theory it should be reverse by date. Is there a `byDate` variant?

carllerche:
I created https://github.com/tokio-rs/website/pull/143 which sorts by weight. This will just require the weight field order to match the date. Not ideal, but it works for now.

I could not find a way to sort by date.

PR #145, opened 2018-03-08
Update for the runtime release

carllerche:
This patch updates the website to use the Tokio runtime instead of
`current_thread`.


carllerche:
This is currently breaking as Mio has not been released to crates.io yet.

PR #150, opened 2018-03-15
Myssing async wrapper

swfsql:


carllerche:
I believe that the `into()` actually wraps the value with `Async`. However, you are correct that this is confusing. Would you mind leaving the explicit `Async` and removing the `into`?

swfsql:
Sure thing! Thanks, I didn't know of the into() effect!

PR #152, opened 2018-03-30
Minor $ tweak for consistency

whitfin:
This is a nothing PR, but I noticed a small inconsistency with the two commands in the hello world example - one using `$` to signal shell and the other not; so this PR just adds the missing `$`.


carllerche:
There is no nothing PRs!

Thanks for takin the time to submit this.

PR #157, opened 2018-04-04
Correct runtime of Interval example

kraai:
Since the Interval is 100 milliseconds and there are 9 such intervals, I think that the total time will be 0.9 seconds, not 9 seconds.


carllerche:
Thanks! Makes a lot more sense now :)

PR #161, opened 2018-04-13
Update docs.rs futures link to use futures@0.1

bryanburgers:
Tokio uses the futures library at futures version 0.1. The sidebar in
the documentation section, however, confusingly pointed to futures
version 0.2. Update the link.

I also checked if there were any other locations where futures was
linked to not using 0.1 explicitely with the following command

    grep -rP 'docs.rs/futures(?!/0.1)'

and changed a legacy document, but everything else looks like it is
explicitely linking to the 0.1 version of the docs.


carllerche:
Thanks!

PR #162, opened 2018-04-15
Describe how one could build one's own runtime

vorner:
Mostly for educational purposes, as a runtime is already provided.

This should not go in before https://github.com/tokio-rs/tokio/pull/306, as it references that.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -6,7 +6,9 @@ googleAnalytics = "UA-90234741-1"
 
 [params]
 tokioDocsURL = "https://docs.rs/tokio/0.1/tokio"
+tokioExecutorDocsURL = "https://docs.rs/tokio-executor/0.1/tokio_executor"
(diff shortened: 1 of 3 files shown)


vorner:
Hmm, any idea why the tests fail? They pass locally and I do have all the `extern crate` crates listed :-(.

carllerche:
@vorner thanks for the PR, I will read it in a bit. To answer the CI question, I believe if you have extern crate you also need a fn main. Other examples hide the main fn but there is one to make ci happy.

vorner:
Yes, that seems to have been it ‒ interestingly, nightly no longer seems to mandate the `fn main`, that's why it passed locally.

I added a fixup (can be squashed before merge). Locally passes with stable as well, so I hope it'll work in CI too.

carllerche:
@vorner Sorry for the delay, I was w/o internet access for the past week.

carllerche:
Looks great to me! I have merged. Thanks 👍

PR #169, opened 2018-05-21
Suggested improvements for the Getting Started guide

cpick:
When I was reading through the Getting Started guide, there were a few
spots where I was a bit confused or felt the wording could be improved.

These are a few suggestions that I think would have helped.


carllerche:
Great! Thanks for providing the PR.

PR #171, opened 2018-05-29
Fix link to mio crate on building a runtime docs page

drklee3:
Just removes an extra `.` at the end of the link.


carllerche:
Thanks!

PR #182, opened 2018-07-10
Update frames.md to fix missing word

mindriot101:
"[...] helpers to transform a stream of bytes into a stream frames" => "[...] helpers to transform a stream of bytes into a stream *of* frames"


mindriot101:
Sorry just realised this is a dupe, my bad...

PR #191, opened 2018-08-28
Fix local website build

### LINKED ISSUE #190: Building the website is broken locally
I'm working on updating the [`impl Trait`](https://tokio.rs/docs/going-deeper/returning/#impl-trait) section of the [returning futures](https://tokio.rs/docs/going-deeper/returning/) page and when I try building the website locally, I get the error:

```
ERROR 2018/08/28 21:59:32 REF_NOT_FOUND: Ref "tasks.md": failed to resolve path from page "/legacy/going-deeper-tokio/core-low-level.md": page reference "tasks.md" is ambiguous
```

[This](https://github.com/tokio-rs/website/blob/master/content/legacy/going-deeper-tokio/core-low-level.md) is the file in question.

bIgBV:
The issue stems from a file in the legacy directory which does not seem to be a part of the current website. I made the link into an absolute link to fix the issue. There are several other warnings regarding the same issue, but not being familiar with Hugo, I couldn't make a lot of headway with it.

But this fix should allow building the website on the latest version of Hugo.

Fixes #190

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,14 @@ language: rust
 sudo: false
 
 install:
-  - curl -L https://github.com/spf13/hugo/releases/download/v0.18.1/hugo_0.18.1_Linux-64bit.tar.gz |
+  - curl -L https://github.com/gohugoio/hugo/releases/download/v0.47.1/hugo_0.47.1_Linux-64bit.tar.gz |
       tar xzvf -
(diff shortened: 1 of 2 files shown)


carllerche:
It looks like this requires a recent hugo. Would you be able to update the CI script to point to the most recent release?

PR #192, opened 2018-08-29
Update to latest version of hugo

bIgBV:
Looks like the `hugo` binary from the latest release just has the name `hugo` on untarring. 

* Made changes to travis.yml to run CI

Blocks #191

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,14 +2,14 @@ language: rust
 sudo: false
 
 install:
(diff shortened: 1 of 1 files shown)


bIgBV:
So looks like in order for the build to pass, the fix from #191 has to be a part of this patch. Should I merge those changes into this branch and make a single PR for both the issues?

bIgBV:
I'll make the related CI changes on the previous PR.

PR #205, opened 2018-09-18
Small Changes to Getting-Started

rylev:
These are small changes to the getting started guide. Mostly I removed information that in my opinion came too early and just made things more confusing. 

Besides these changes I have some thoughts/suggestions:
* It would be better to introduce Futures and Tasks purely through combinators and only once the reader has a good understanding of that, introduce writing custom Futures. 
* Now that `impl Trait` is stable, it might make sense to use that syntax everywhere and just make a short mention that if the reader wants to use an older version of Rust, they should look at the "Returning Futures" guide.
* `ErrorKind::WouldBlock` is mentioned a few times, but the underlying mechanism for resources to go from `NotReady` to `Ready`, I wonder if hand waving and never even mentioning `WouldBlock` would be less confusing.

### DIFF
--- a/content/docs/getting-started/futures.md
+++ b/content/docs/getting-started/futures.md
@@ -108,10 +108,6 @@ trait Future {
 }
 ```
 
-You may notice that this is the exact same trait that was used to implement an
-asynchronous task. This is because asynchronous tasks are "just" futures that
-complete with a value of `()` once the computation has completed.
-
 Usually, when you implement a `Future`, you will be defining a computation that
 is a composition of sub (or inner) futures. In this case, the future implementation tries
 to call the inner future(s) and returns `NotReady` if the inner futures are not
@@ -183,13 +179,12 @@ fn poll(&mut self) -> Result<Async<usize>, T::Error> {
 
 # Returning `NotReady`
 
-The last section handwaved a bit and said that when a task returns `NotReady`,
-once it transitioned to the ready state, the executor is notifed. This enables
-the executor to be efficient in scheduling tasks.
+The last section handwaved a bit and said that once a Future transitioned to the
+ready state, the executor is notifed. This enables the executor to be efficient
+in scheduling tasks.
 
-When a function returns `Async::NotReady`, it is critical that the executor is
-notified when the state transitions to "ready". Otherwise, the task will hang
-infinitely, never getting run again.
+It is critical that the executor is notified when the state transitions to "ready".
+Otherwise, the task will hang infinitely, never getting run again.
 
 For most future implementations, this is done transitively. When a future
 implementation is a combination of sub futures, the outer future only returns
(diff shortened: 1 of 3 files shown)


content/docs/getting-started/futures.md
```diff
@@ -183,13 +179,12 @@ fn poll(&mut self) -> Result<Async<usize>, T::Error> {
...
-once it transitioned to the ready state, the executor is notifed. This enables
-the executor to be efficient in scheduling tasks.
+The last section handwaved a bit and said that once a Future transitioned to the
+ready state, the executor is notifed. This enables the executor to be efficient
+in scheduling tasks.
 
-When a function returns `Async::NotReady`, it is critical that the executor is
-notified when the state transitions to "ready". Otherwise, the task will hang
-infinitely, never getting run again.
+It is critical that the executor is notified when the state transitions to "ready".
```
carllerche: It is only critical that the executor is notified if the function previously returned `NotReady`.

Perhaps:

> When a function returns `Async::NotReady`, it signals that it is currently not in a ready state and is unable to complete the operation. It is critical that the executor is notified when the state transitions to "ready". Otherwise, the task will hang infinitely, never getting run again.

rylev: This sounds better to me. I like your suggestion!

rylev:
@carllerche any thoughts on the bigger suggestions/comments I made in the description of this PR?

carllerche:
> It would be better to introduce Futures and Tasks purely through combinators and only once the reader has a good understanding of that, introduce writing custom Futures.

I'm not sure about this one. IMO implementing `Future` and using combinators are equals right now. This will change w/ `async` / `await`, but right now users should not feel like implementing `Future` is special.

> Now that `impl Trait` is stable, it might make sense to use that syntax everywhere and just make a short mention that if the reader wants to use an older version of Rust, they should look at the "Returning Futures" guide.

Yeah, probably does. You still want to know about other options for cases where you need to name the type.

> `ErrorKind::WouldBlock` is mentioned a few times, but the underlying mechanism for resources to go from `NotReady` to `Ready`, I wonder if hand waving and never even mentioning `WouldBlock` would be less confusing.

Sounds like a good idea just reading it. Each instance would have to be evaluated, but odds are it is fine to not mention `WouldBlock` (except maybe in the internals guide).

rylev:
@carllerche for your second and third comments I'll submit new PRs with the changes. 

As for your first comment: in my head I think I can make it work without de-emphasizing the custom future. It would purely be an attempt to rearrange the content so that more accessible things are introduced first and then more complex things. I won't do anything in this PR, but I'll come up with a proposal in another PR.

rylev:
@carllerche I took your suggestion so this should be ready for merge now!

PR #209, opened 2018-09-21
Change how and when ErrorKind::WouldBlock is mentioned

rylev:
In general `ErrorKind::WouldBlock` is an implementation detail and not super relevant to the topic at hand. It was mentioned twice, once of which I completely removed. The other mention I still left in, but de-emphasized it by moving it into parentheses.

### DIFF
--- a/content/docs/getting-started/io.md
+++ b/content/docs/getting-started/io.md
@@ -39,9 +39,9 @@ mandated properties:
 
 * Calls to `read` or `write` are **nonblocking**, they never block the calling
   thread.
-* If a call would otherwise block then an error is returned with the kind of
-  `WouldBlock`. If this happens then the current future's task is scheduled to
-  receive a notification (an unpark) when the I/O is ready again.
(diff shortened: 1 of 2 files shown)


content/docs/getting-started/io.md
```diff
@@ -39,9 +39,9 @@ mandated properties:
 
 * Calls to `read` or `write` are **nonblocking**, they never block the calling
   thread.
-* If a call would otherwise block then an error is returned with the kind of
-  `WouldBlock`. If this happens then the current future's task is scheduled to
-  receive a notification (an unpark) when the I/O is ready again.
+* If a call would otherwise block then the function returns a value indicating so.
+  If this happens then the current future's task is scheduled to receive a
+  notification (known as an 'unpark') when the I/O is ready again.
```
carllerche: The unpark terminology has been abandoned in favor of just calling it a task notification.

Not in this PR, I would greatly appreciate a scrub for “unpark” if you are able to :-)

PR #217, opened 2018-10-02
Try to at least briefly explain new terms when they are introduced

rylev:


content/docs/overview.md
```diff
@@ -76,19 +76,7 @@ such as accessing uninitialized memory and use after free. It does this without
...
-
-# let foo = "".to_string();
-let data = Mutex::new(foo);
-
-let locked = data.lock().unwrap();
-println!("locked data: {}", &locked[..]);
-
-// The lock is automatically released here when `locked` goes out of scope.
-```
+APIs. For example, `Mutex` does not require the user to explicitly unlock.
```
rylev: I don't believe a full code example about Mutex's is appropriate here. Maybe linking to the std-lib docs.

PR #223, opened 2018-10-05
Simplify the hello world Example

rylev:
The current hello-world example is fairly complex. It uses Streams and Futures, and it spawns sub-tasks on the runtime from within a combinator. It also uses sockets which might require a bit more background knowledge than files do. I think ultimately the "example" page of the getting started section should be as complex as the current hello-world example. 

The hello world example is now truely minimal. It has two futures and it spawns the parent future on the runtime. It deals with files which everyone should be familiar with.

### DIFF
--- a/content/docs/getting-started/hello-world.md
+++ b/content/docs/getting-started/hello-world.md
@@ -7,10 +7,14 @@ menu:
 ---
 
 To kick off our tour of Tokio, we will start with the obligatory "hello world"
-example. This program will create a file and write "hello, world!" to the file.
-The difference between this and a Rust program that writes to a file without Tokio
-is that this program won't block program execution when the file is created or when
-our "hello, world!" message is written to the file.
+example. This program will create a TCP stream and write "hello, world!" to the stream.
+The difference between this and a Rust program that writes to a TCP stream without Tokio
+is that this program won't block program execution when the stream is created or when
+our "hello, world!" message is written to the stream.
+
+Before we begin you should have a very basic understanding of how TCP streams work. Having
+an understanding of Rust’s [standard library implementation](https://doc.rust-lang.org/std/net/struct.TcpStream.html)
+is also helpful.
 
 Let's get started.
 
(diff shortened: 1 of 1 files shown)


content/docs/getting-started/hello-world.md
```diff
@@ -37,48 +35,42 @@ and the crates and types into scope in `main.rs`:
...
 use tokio::prelude::*;
 # fn main() {}
 ```
 
-# Writing the server
+# Creating the file
 
-The first step is to bind a `TcpListener` to a local port. We use the
-`TcpListener` implementation provided by Tokio.
+The first step is to create the `File`. We use the `File` implementation provided by Tokio.
```
mariosangiorgio: Could it be worth mentioning what the difference is between `std::fs::File` and `tokio::fs::File`?

I'd follow what the [documentation](https://docs.rs/tokio/0.1.11/tokio/fs/file/struct.File.html) says.

> The first step is to create the `File`. We use the `File` implementation provided by Tokio. This is a specialized version of `std::fs::File` for usage from the Tokio runtime.

It's probably worth pointing out that once we have a`tokio::fs::File` we will have to interact with it with tokio-specific functions. E.g. [`tokio::io::Write::write_all`](https://tokio-rs.github.io/tokio/tokio/io/trait.Write.html#method.write_all) vs [`std::io::Write::write_all`](https://doc.rust-lang.org/std/io/trait.Write.html#method.write_all).

carllerche:
Thanks for doing this. I agree that minimizing the "hello world" example would be good.

I hesitate to switch to using files as the example because Tokio is _primarily_ for asynchronous networking. Tokio's file support is mostly to provide file access while you are performing network operations (send the contents of a file to a socket).

Also, files don't go via epoll / kqueue / etc... Anybody who does try to look at Tokio w/ that prior knowledge might end up more confused.

What if, instead of implementing a server for the hello world example, a TCP client is used instead? It should mirror the file pretty closely.

Instead of `File::create` you would have `TcpStream::connect`. You could also use the `write_all(...)` call.

Thoughts?

rylev:
I switched to using TcpStream instead of File. Let me know what you think. Once this is approved, I can change the hello world example in the Tokio repo.

carllerche:
As a note, server examples should come back in the next section: https://github.com/tokio-rs/doc-push/blob/master/outline/io-with-tokio.md

PR #224, opened 2018-10-06
Rework "Getting Started"

rylev:
This is work in progress!

Implements: https://github.com/tokio-rs/doc-push/blob/master/outline/getting-started.md

So far I've left out the error handling page. The more I think about it, the more it seems like error handling would be the wrong topic here. I believe it deserves its own section somewhere else in the guide. The responsibility of the Getting Started section is really just to give the reader a high level overview of how to use Tokio and a basic conceptual framework of how Tokio works.

I've written only a little bit myself. Most is just a reordering of the Futures, Tasks, and Runtime pages as well as significant cuts to things that will be handled in the next section on Futures, Sinks and Streams as well as other sections. 

This part of the guide will no longer cover:
* Message passing
* Implementing a future from scratch
* A naive executor

These topics are left for later in the guide.


carllerche:
I skimmed over, and the direction looks good (I will look in more detail now).

However, I would feel more comfortable removing major sections (like tasks) once a replacement has been put into place. As of now, this is the only source for the documentation.

rylev:
@carllerche I definitely understand. I think the next thing to do is to lay down the scaffolding for the next two major sections and move the deleted material into those sections. Then we at least haven't lost anything

content/docs/getting-started/futures.md
```diff
@@ -1,19 +1,72 @@
...
 
-# What Are Futures?
+We'll be taking a closer look at the runtime in the next section, but a basic
+understanding of the runtime is necessary to understand futures. To gain this
+understanding, we'll firs tlook at the synchronous model that Rust uses by default
+and see how this differs from Tokio's asynchronous model.
+
+# Synchronous Model
+
+First, let's talk briefly about the synchronous (or blocking) model. This is the
```
carllerche: Probably reword to:

> "The Rust [standard library] uses a synchronous (or blocking) model.

The section above has "we'll first look at the synchronous model that Rust uses by default", so it is repeated.

content/docs/getting-started/futures.md
```diff
@@ -1,19 +1,72 @@
...
+// let socket = ...;
+let mut buf = [0; 1024];
+let n = socket.read(&mut buf).unwrap();
+
+// Do something with &buf[..n];
+# }
+```
+
+When `socket.read` is called, either the socket has pending data in its receive
+buffer or it does not. If there is pending data, then the call to `read` will
```
carllerche: ~~then~~ the call to `read` will

content/docs/getting-started/futures.md
```diff
@@ -1,19 +1,72 @@
...
+return immediately and `buf` will be filled with that data. However, if there is
+no pending data, then the `read` function will block the current thread until
+data is received. At which time, `buf` will be filled with this newly received
```
carllerche: "At which time" reads wrong to me.

Maybe repeat "Once the data is received, "

content/docs/getting-started/futures.md
```diff
@@ -1,19 +1,72 @@
...
+data and the `read` function will return.
+
+In order to perform reads on many different sockets concurrently, a thread per
+socket is required. Using a thread per socket does not scale up very well to
+large numbers of sockets. This is known as the [c10k] problem.
+
+# Non-blocking sockets
+
+The way to avoid blocking a thread when performing an operation like read is to
```
carllerche: "The way to avoid blocking a thread when performing an operation like read is to not block the thread!"

Probably can be phrased better!

"Non-blocking sockets allow performing operations, like read, without blocking the thread".

content/docs/getting-started/futures.md
```diff
@@ -57,8 +110,8 @@ let response_future = client.get("https://www.example.com");
 Now, the `response_future` isn't the actual response. It is a future that will
 complete once the response is received. However, since the caller has a concrete
 **thing** (the future), they can start to use it. For example, they may chain
```
carllerche: Maybe sub **thing** with **value**.

content/docs/getting-started/futures.md
```diff
@@ -74,282 +127,105 @@ They cannot because they don't have the actual HTTP response. Instead, they
...
-# Implementing `Future`
+We'll be exploring more combinators later in this guide.
 
-Implementing the `Future` is pretty common when using Tokio, so it is important
-to be comfortable with it.
+# Poll based Futures
 
-As discussed in the previous section, Rust futures are poll based. This is a
-unique aspect of the Rust future library. Most future libraries for other
+As discussed in the previous section, Rust futures are poll based. This means that
```
carllerche: This is no longer discussed in the previous section.

content/docs/getting-started/futures.md
```diff
@@ -74,282 +127,105 @@ They cannot because they don't have the actual HTTP response. Instead, they
...
+future. For instance:
+* **UI Events** caused by the user interacting with a GUI in different ways. When an
+  event happens the stream yields a different message to your app over time.
+* **Push Notifications from a server**. Sometimes a request/response model is not
+  what you need. A client can establish a notification stream with a server to be
+  able to receive messages from the server without specifically being requested.
+* **Incoming socket connections**. As different clients connect to a server, the
+  connections stream will yield socket connections.
+
+Unlike futures, streams never yield values when they're complete, but rather
```
carllerche: This sentence is a bit confusing. Should it be removed? The section's opening paragraph covers what is needed I think?

content/docs/getting-started/futures.md
```diff
@@ -74,282 +127,105 @@ They cannot because they don't have the actual HTTP response. Instead, they
 define the work to be done when the response future completes.
 
 Both the [`futures`] crate and Tokio come with a collection of combinator
-functions that can be used to work with futures.
+functions that can be used to work with futures. So far we've seen `and_then` which
+chains two [`futures`] together, `then` which allows to chain a future to a previous
+one even if the previous one errored, and `map` which simply maps a future's value
+from one type to another.
```
carllerche: It might be worth adding some code examples w/ some other combinators? Maybe showing select and join?

content/docs/getting-started/futures.md
```diff
@@ -74,282 +127,105 @@ They cannot because they don't have the actual HTTP response. Instead, they
...
+about `poll` in this section since you don't need to know about `poll` to use futures
+with combinators. The only thing to be aware for now is that `poll` is
+what the runtime will call in order to see if the `Future` is complete yet or not. If
+you're curious: `Async` is an enum with values `Ready(Item)` or `NotReady` which
+informs the runtime of if the future is complete or not.
+
+In a future section, we'll be implementing a `Future` from scratch including writing
+a poll function that properly informs the runtime when the future is complete.
+
+## Streams
```
carllerche: I think this section could use a bit on stream combinators to show how they differ from futures. I would show `map` (because `Future::map` was mentioned earlier), but also `for_each` which returns a future.

content/docs/getting-started/runtime.md
```diff
@@ -0,0 +1,94 @@
...
+a value (in the case of a `Future`) or a series of values (in the case of `Stream`)
+that will be available "at some point in the future". We talked about `poll` on `Future`
+and `Stream` which the runtime will call to figure out if the `Future` or the
+`Stream` are ready to yield a value.
+
+Lastly, we hinted at the idea of an exectuor which is the specific thing inside the
+runtime which polls Futures and Streams driving them to completion. We'll take a
+closer look at executors now.
+
+## Executors
```
carllerche: In the getting started part, maybe we shouldn't use the "executor" term and instead focus on "runtime".

goriunov:
This is just my opinion.

I have noticed that  `getting-started/futures.md` has a `Synchronous Model` example with `read` method and because of that guide seems to be a bit incomplete because i can not see any alternative to the `Synchronous Model` example.  It could be the same example implemented with Futures some where in the Futures section.

I am not good at tokio yet so i will give simple abstract of what i mean :)

***Everything below is just and example***

# Synchronous Model
First, let's talk briefly about the synchronous (or blocking) model. ....
```
Synchronous code example
```

# Poll based Futures
As discussed in the previous section, Rust futures are poll based. This means that ......
This is how `Synchronous Model`(above) example will look if you use  Poll based Futures:
```
Example of Synchronous code transformed to Futures .
```
**some explanation**
More details will be explained in the next section ...

content/docs/getting-started/futures.md
rylev: Hmm yea that could be nice, but I don't want this to turn into a guide on combinators. Let me play with it a bit.

AndyGauge:
### Getting Started / Hello World

I enjoyed creating the program, and then I ran it to find 
> connection error = Os { code: 111, kind: ConnectionRefused, message: "Connection refused" }
and
> connection error = Os { code: 10061, kind: ConnectionRefused, message: "No connection could be made because the target machine actively refused it." }

These are perfectly normal, and I realized I didn't have a server running.  A novice running a hello world example will think they did something wrong (I did).  Can we say in the description following the tokio::run(client); line that AS-IS this code will return an error unless the server is running.  The server code can be found [here](https://github.com/tokio-rs/tokio/blob/master/examples/hello_world.rs#L1)

tobz:
I think I'd rather we talk the user through starting the server component.  If the goal is to give a step-by-step, then running the server will always be a step, and we should be telling them _that_ rather than (likely) sending them down the unhappy path first and giving them the information to figure out why they're there.

rylev:
@tobz @AndyGauge #233 should hopefully fix issues with the hello world example.

carllerche:
@rylev a thought. If you want to get this merged sooner than later, the deleted guides can be moved temporarily to "Going deeper".

rylev:
@carllerche sorry this took so long. I think I finally got it to a place where it's mergable (no information is lost). I think we can save the other improvements for future pull requests.

PR #231, opened 2018-10-11
IO Overview WIP

AndyGauge:
[LIVE](https://www.yetanother.site/website/docs/io/overview/)

The IO Overview will include:

> Functions as an introduction / tutorial to using I/O with Tokio.
> Comes w/ TCP, UDP, Unix
> Non-blocking
> Backed by Mio
> based on runtime model (previous section).
> Setup a TCP server
> Connect with TcpStream
> Send some data using io::write_all
> Read data on server, print, and close socket (note on drop).

[doc-push outline](https://github.com/tokio-rs/doc-push/blob/master/outline/io-with-tokio.md#overview)

### DIFF
--- a/content/docs/io/overview.md
+++ b/content/docs/io/overview.md
@@ -53,13 +53,29 @@ the buffer, and reevaluating `.incoming()`.
 allow an event loop to respond to I/O rather than wait for it.  That event loop
 is [`Tokio::reactor`] which receives events from [`mio`].
 
+[`mio::net`] implements networking primitives that allow polling events.
+[`mio::unix`] implement UNIX inter-process communication channels with polling.
+
+## Included with Tokio
+
+Tokio provides the high level interfaces to asynchronous IO for
+
+  * [`TCP sockets`]
+  * [`UDP sockets`]
+  * [`Unix sockets`]
+
 
 [`AsyncRead`]: {{< api-url "tokio" >}}/io/trait.AsyncRead.html
 [`AsyncWrite`]: {{< api-url "tokio" >}}/io/trait.AsyncWrite.html
 [`futures::Async`]: {{< api-url "futures" >}}/enum.Async.html
-[`mio`]: https://docs.rs/mio/*/mio/
+[`mio`]: {{< api-url "mio" >}}
+[`mio::net`]: {{< api-url "mio" >}}/net/index.html
+[`mio::unix`]: {{< api-url "mio" >}}/unix/index.html
 [`std::fs`]: https://doc.rust-lang.org/std/fs/struct.File.html#implementations
 [`std::io`]: https://doc.rust-lang.org/std/io/#read-and-write
 [`std::net` examples]: https://doc.rust-lang.org/std/net/struct.TcpListener.html#examples
+[`TCP sockets`]: {{< api-url "tokio" >}}/net/tcp/index.html
 [`tokio`]: {{< api-url "tokio" >}}
 [`Tokio::reactor`]: {{< api-url "tokio-reactor" >}}
+[`UDP sockets`]: {{< api-url "tokio" >}}/net/udp/index.html
+[`Unix sockets`]: {{< api-url "tokio" >}}/net/unix/index.html
--- a/content/docs/io/overview.md
+++ b/content/docs/io/overview.md
@@ -15,9 +15,44 @@ The [`tokio`] crate comes with a variation on those types, [`AsyncRead`] and
 is the [`futures::Async`] return values.  The implementation returns an `Error`,
 `NotReady`, or `Ready` value.
 
+## Non Blocking
+
+The nature of a polling data reader/writer allows for other tasks to operate
+before data reaches the socket or disk queue.  Synchronous operations evaluate
+in a guaranteed sequential manner, and require commands to complete before the
+next operation executes.
+
+Consider this `std::net` implementation from the [`std::net` examples]
+
+```rust,no_run
+# #![allow(unused)]
+# use std::io;
+use std::net::{TcpListener, TcpStream};
+
+# fn handle_client(stream: TcpStream) {
+#     // ...
+# }
+#
+fn main() -> io::Result<()> {
+    let listener = TcpListener::bind("127.0.0.1:8000")?;
+#
+#     // accept connections and process them serially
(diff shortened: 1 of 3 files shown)


content/docs/io/overview.md
```diff
@@ -6,7 +6,109 @@ menu:
...
+files and sockets.  These `Read` and `Write` traits are implemented in other
+modules including [`std::fs`] and `std::net`.
 
-[doc-push]: https://github.com/tokio-rs/doc-push
+The [`tokio`] crate comes with a variation on those types, [`AsyncRead`] and
+[`AsyncWrite`].  The primary difference between [`AsyncRead`] and `std::io::Read`
+is the [`futures::Async`] return values.  The implementation returns an `Error`,
+`NotReady`, or `Ready` value.
+
+## Non Blocking
```
carllerche: We might want to name this section "Blocking vs. Non-Blocking" or something like that.

You can also look at [Node's page](https://nodejs.org/en/docs/guides/blocking-vs-non-blocking/) for inspiration.

content/docs/io/overview.md
```diff
@@ -6,7 +6,109 @@ menu:
...
+
+The nature of a polling data reader/writer allows for other tasks to operate
```
carllerche: I would stick with "non-blocking" as the term (vs. polling).

"Non-blocking reading and writing allows for other tasks to operate concurrently" maybe?

content/docs/io/overview.md
```diff
@@ -6,7 +6,109 @@ menu:
...
+before data reaches the socket or disk queue.  Synchronous operations evaluate
```
carllerche: Can we expand more about how non-blocking allows concurrent tasks? And break the synchronous / blocking counterpoint to a new paragraph.

content/docs/io/overview.md
```diff
@@ -6,7 +6,109 @@ menu:
...
+        handle_client(stream?);
+    }
+#    Ok(())
+}
+```
+
+This single-threaded process will evaluate through calling `.incoming()` on
+the `TcpListener` and will continue to run until receiving a KILL signal from
+the operating system or until enough data is received on the TCP port to flush
+the buffer, and reevaluating `.incoming()`.
```
carllerche: Maybe further expand by saying that it is impossible to run other logic on the thread while blocked on the `incoming()` call. In order to process the client while concurrently accepting new sockets, a thread must be spawned (which has overhead).

content/docs/io/overview.md
```diff
@@ -6,7 +6,109 @@ menu:
...
+
+`Tokio` I/O, rather than pausing execution, uses a `Futures` based syntax to
```
carllerche: "**non-blocking** `Futures` based syntax" (trying to keep the same terminology throughout.

carllerche: "`Tokio` I/O, rather than pausing execution **of the thread**"?

content/docs/io/overview.md
```diff
@@ -6,7 +6,109 @@ menu:
...
+allow an event loop to respond to I/O rather than wait for it.  For network I/O
+the driver is [`Tokio::reactor`] which receives events from [`mio`].
+
+### Mio
```
carllerche: After reading this, I would probably get rid of the Mio sections. It doesn't seem relevant to users trying to get started w/ Tokio & I/O. Sorry about that. 

That said, some of the content would be useful at the appropriate locations in other pages.

content/docs/io/overview.md
```diff
@@ -6,7 +6,109 @@ menu:
...
+
+Mio is a [cross platform] implementation of low-level I/O.  Mio's goal is to
+create a unified syntax to operate on a wide range of platforms.  This makes
+Tokio applications portable by default.
+
+#### Evented
+
+Mio includes a trait [`Evented`] which designates a source of `Event`.  The
```
carllerche: I wouldn't mention `Evented` here as it is an implementation detail. I believe it is introduced in a later section.

content/docs/io/overview.md
```diff
@@ -6,7 +6,109 @@ menu:
...
+[`mio::net`] implements networking primitives that allow polling events.
+[`mio::unix`] implement UNIX inter-process communication channels with polling.
+
+### Tokio Runtime
+
+`mio` does not provide scheduling.  Tokio shines by using configuration to
+implement a `Runtime Model` that matches the properties of a specific resource.
+
+
+## Included with Tokio
```
carllerche: I recommended including this content at the top.

content/docs/io/overview.md
```diff
@@ -6,7 +6,109 @@ menu:
...
+Tokio applications portable by default.
+
+#### Evented
+
+Mio includes a trait [`Evented`] which designates a source of `Event`.  The
+implementation of the trait [`register`]s with [`Poll`].  To receive an `Event`,
+the executor will `poll` the registered `Evented` value and forward it to the
+pending future task.
+
+#### Polling
```
carllerche: I think this info might be more relevant in the "internals" section. I am not sure if it is included today, if not we should add it.

content/docs/io/overview.md
```diff
@@ -6,7 +6,109 @@ menu:
...
+}
+```
+
+This single-threaded process will evaluate through calling `.incoming()` on
+the `TcpListener` and will continue to run until receiving a KILL signal from
+the operating system or until enough data is received on the TCP port to flush
+the buffer, and reevaluating `.incoming()`.
+
+`Tokio` I/O, rather than pausing execution, uses a `Futures` based syntax to
+allow an event loop to respond to I/O rather than wait for it.  For network I/O
```
carllerche: Maybe something like:

> [...] based syntax allowing tasks to respond to I/O events as they arrive rather than wait for them. Networking events are provided by `mio`, a non-blocking I/O library backed by the operating system's event queue (`epoll`, `kqueue`, `iocp`, ...).

AndyGauge:
After getting this far, I had felt similar sentiments about the progress.  I guess most of this was the background required to understand Tokio personally.  I appreciate the direction and the specific guidance.

carllerche:
I'm going to close this PR as #263 has continued on this work.

Thanks so much @AndyGauge for getting this started <3.

PR #232, opened 2018-10-13
Be more explicit on how to include crates maybe?

jasikpark:
Hopefully anyone who is trying out the tokio tutorial has worked through *the book* before, but if they haven't and they haven't included crates before, I could imagine not referencing the specific file to be a problem.


jasikpark:
This PR is also referenced in #233

PR #234, opened 2018-10-14
Improve hello world code consistency - `stream` and `hello_world` should be `client`

jasikpark:
On the `hello_world` example, there are three different variables used, all for the same variable: `stream`, `hello_world`, and `client`.

In the provided [completed example](https://github.com/tokio-rs/tokio/blob/master/examples/hello_world.rs), the variable `client` is used.

This is a pull request to simply set all of the different variables to `client` and be less confusing.


jasikpark:
This is referenced in #224, but I thought it would a good idea to pull this out into it's own PR, to make it easy to merge.

jasikpark:
I had Img-Bot trying to losslessly optimize all images in all of my repos, so that's where all the extra commits came from. I'm not entirely sure how to not include them/`git rebase` them

rylev:
@jasikpark thanks for this! I think it would be good to remove the imgbot commits for now - If you think it's worth using, you can create an issue. Maybe, closing this PR and opening a new one without the imgbot changes?

jasikpark:
right, ok

jasikpark:
the new pull request is #235

PR #235, opened 2018-10-14
[New] Improve hello-world consistency - hello_world and stream should be client

jasikpark:
Rename `hello_world`, `stream`, and `client` to `client`.

A cleaned up PR of #234


carllerche:
This seems good to me, but I will let @rylev approve / merge. Thanks for the PR 👍

Issue #230, opened 2018-10-11
Stop using hard line breaks


rylev:
The documentation markdown files have a ~80 column limit. While this might be nice for editing, it makes the pages themselves look terrible. We should let the page itself decide when newlines happen.

carllerche:
@rylev could you clarify how this makes the pages look terrible? The line break in MD should not carry over to the rendered content.

jasikpark:
If we *did* want to only use single-line paragraphs, [remark](https://remark.js.org) seems to be a good library/cli for linting markdown and for automatically formatting markdown. Even without the decision to change how linebreaks are done, it still could be useful for it's other style formatting and linting features.

jasikpark:
Another option would be [comrak](https://github.com/kivikakk/comrak), a rust library that does the same.

carllerche:
I'm still not sure that I understand the motivation. What is the problem being solved?

jasikpark:
I don't think that there's a problem being solved by requiring single-line paragraphs, but by having the markdown documents all formatted by an autoformatter, the structure can be easier to see and manage. But it's not necessarily needed.

rylev:
The original reason I opened this issue was because I mistakenly thought the width of the rendered text was contrained by the line breaks in the markdown files (instead of being determined by CSS). I personally think the text width is too small, but this does not have to do with newlines in the markdown files, so I'm closing. @jasikpark If you think having a markdown formatter run over the markdown files, please open another issue. I will open one regarding the width of the text.

PR #236, opened 2018-10-16
Change to soft-linebreaks

barskern:
This changes every markdown file (except `README.md`) to use soft instead of hard linebreaks (e.g all paragraphs are in one line). I saw in the issue that @rylev specified "documentation markdown files", however its a simple change to remove files from this PR.

PS! Going through the docs I found what seemed like a [invalid link](https://github.com/tokio-rs/website/compare/master...barskern:chore/change-to-soft-linebreaks?expand=1#diff-e806a3a3b56e46e79847fccefef92337R380) which I annotated with 'FIXME' 😄 

Closes: #230

### DIFF
--- a/content/blog/2017-01-tokio-0-1.md
+++ b/content/blog/2017-01-tokio-0-1.md
@@ -8,50 +8,22 @@ weight = 999
 
 Today we are publishing the preliminary version of the Tokio stack, 0.1!
 
-**Tokio is a platform for writing fast networking code in Rust.** It's built on
-futures,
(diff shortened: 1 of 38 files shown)


carllerche:
Thanks for doing this. Before merging, I'd like to get a bit more discussion in #230 first. As mentioned, getting rid of the line break makes editing more difficult, so maybe there is a way to fix this without removing it.

barskern:
Yeah thats a good point, at least now its ready if you end up wanting it 😄

carllerche:
@barskern thanks for doing this work even though it isn’t getting merged :heart:

barskern:
No problem 😊

PR #239, opened 2018-10-29
Some very minor changes to the first chapters

nrc:
Some grammar fixes, some indentation fixes to make it easier to copy and paste code, and a few changes to try and make things clearer.

### DIFF
--- a/content/docs/getting-started/hello-world.md
+++ b/content/docs/getting-started/hello-world.md
@@ -21,7 +21,7 @@ Let's get started.
 First, generate a new crate.
 
 ```bash
-$ cargo new --bin hello-world
+$ cargo new hello-world
 $ cd hello-world
 ```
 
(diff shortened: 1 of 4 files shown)


content/docs/getting-started/hello-world.md
```diff
@@ -21,7 +21,7 @@ Let's get started.
 First, generate a new crate.
 
 ```bash
-$ cargo new --bin hello-world
```
nrc: `--bin` is the default for quite a while now

content/docs/getting-started/hello-world.md
```diff
@@ -203,7 +203,7 @@ You can find the full example [here][full-code].
 command starts a listening TCP socket on the previously specified port.
 
 ```bash
-$ nc -l -p 6142
+$ nc -l 6142
```
nrc: I'm not 100% sure about this, but on Mac OS at least, `-p` causes an error and netcat works without it. I haven't tested on other platforms but this does correspond to some documentation I found.

carllerche:
Thanks, these look good.

Unfortunately, there are conflicts. Instead of rebasing, I'll try including these changes as part of #241.

rylev:
@carllerche @nrc Should we close this? Has this been integrated into #241?

carllerche:
I need to double check...

carllerche:
These changes were applied in #263

Thanks for the original PR <3

PR #247, opened 2018-11-20
Clarify Box<Future<...> + Send> (doc-push #87)

jeff-at-dwelo:
This explicitly adds the Send trait to all boxed Future examples, and adds
some language to explain why. Also added another example of when Box is
required (trait definitions). Documentation on `impl Trait` is unfortunately sparse because it's a new feature in Rust, and having to box a return from a trait definition is where I personally ran into problems.

I strongly feel that `+ Send` should be the default for all boxed future examples, because forgetting it leads to serious headaches. (Honestly? We should patch futures to impl Send. But I'm new here.) Generally any example folks might copy-paste from should "just work". The rustdoc says to use `tokio::run` for handling execution in most cases, and by default Send is a requirement there.

### DIFF
--- a/content/docs/futures/combinators.md
+++ b/content/docs/futures/combinators.md
@@ -548,7 +548,7 @@ Another strategy is to return a boxed future as a [trait object]:
 # use std::io;
 # use futures::Future;
 # fn main() {}
-fn foo() -> Box<Future<Item = u32, Error = io::Error>> {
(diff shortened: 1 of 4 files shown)


carllerche:
@jeff-at-dwelo Thanks for doing this, this should hopefully avoid some future confusion (pun intended).

PR #250, opened 2018-11-25
Remove extra # in stream example

bobrik:
I think this is how was intended to look like.


rylev:
@bobrik I believe the other instance actually is the opposite problem. Those lines should be completely hidden from the user but are not.

PR #260, opened 2018-11-26
Write the first few sections of I/O with Tokio

jonhoo:
This builds somewhat on #231 by @AndyGauge for the overview. More sections to come.


content/docs/io/reading_writing_data.md
```diff
@@ -6,7 +6,156 @@ menu:
...
+> Note: in this section, we'll primarily talk about `AsyncRead`, but
+> `AsyncWrite` is pretty much exactly the same, just for writing data to
+> an I/O resource (like a TCP socket) instead of reading from it.
+
+So, let's take a look at [`AsyncRead`] and see what all the fuzz is
+about:
+
+```rust,no_run
+use std::io::Read;
+pub trait AsyncRead: Read {
```
carllerche: `AsyncRead` having `std::io::Read` as a super trait is a mistake IMO. Unfortunately, we can't deprecate it w/o a breaking change. This will change in 0.2.

What do you think about omitting `std::io::Read` here and only just bring it up when needed? I personally am not sure what the right thing to do here is. I'm leaning towards **sticking with what you have**, but i thought that I would bring up the thought.

jonhoo:
I'd like to push a little bit back on your comment about code examples. Examples are good, but can also just confuse if introduced too early. The "Overview" section is more about "What even is async I/O?", and I think throwing code at the reader at that point will probably just make them try to read the code instead of understand the model. The exception is the heavily-documented code segment at the bottom of the Overview, which gives a complete, self-contained example that walks the reader through exactly what's going on, and how it ties back to the text above.

I *do* agree with you though that examples would be a good addition to the "Working with I/O futures" section of "Reading and Writing data", as well as in later sections (we'll probably want one on interactions between `AsyncRead/Write` and `Stream/Sink` for example).

carllerche:
@jonhoo Oh, overview is fine... I was thinking about "Reading and Writing Data". It talks about `AsyncRead` but doesn't really show how to use it.

content/docs/io/reading_writing_data.md
carllerche: I'm fine w/ this if you like it. I just wanted to point out that the super trait is going away

jonhoo:
I was imagining that the next section from the overview, "[`poll_` API](https://github.com/tokio-rs/doc-push/blob/master/outline/io-with-tokio.md#poll_-api)", would go into more detail about using `AsyncRead` and `AsyncWrite`, since the overview for "[Reading and Writing Data](https://github.com/tokio-rs/doc-push/blob/master/outline/io-with-tokio.md#reading-and-writing-data)" seemed to focus more on what `AsyncRead` and `AsyncWrite` _are_, and how they're used with combinators. But I'm also happy to move of the discussion of `poll_*` into "Reading and Writing Data".

content/docs/io/overview.md
```diff
@@ -6,7 +6,119 @@ menu:
     parent: io
 ---
 
-This page has not been worked on yet. If you'd like to contribute visit the [doc-push]
-repo.
+The Rust standard library provides support for networking and I/O, such
+as TCP connections, UDP sockets, reading from and writing to files, etc.
+However, those operations are all synchronous, or _blocking_, meaning
+that when you call them, the current thread may stop executing and go to
+sleep until it is unblocked. For example, `read` will block until there
```
rylev: nit: should you mention what `read` does? Where does it read information? If I've never heard of `read` before will be I confused by this?

content/docs/io/overview.md
```diff
@@ -6,7 +6,119 @@ menu:
...
+To enable this, Tokio provides _non-blocking_ versions of many standard
+library I/O resources, such as [file operations] and [TCP], [UDP], and
+[Unix] sockets. They return futures for long-running operations (like
+accepting a new TCP connection), and implement non-blocking variants of
+`std::io::Read` and `std::io::Write` called `AsyncRead` and
+`AsyncWrite`.
+
+Non-blocking reads and writes do not block if, for example, there is no
+more data available. Instead, they return immediately with a
+`WouldBlock` error, along with a promise (like `Future::poll`) that they
```
rylev: I find the use of the word "promise" here to be potentially problematic. I only understand what you mean because I know the implementation. These calls don't _return_ a promise (which seems to be implied here).

content/docs/io/overview.md
```diff
@@ -6,7 +6,119 @@ menu:
...
+have arranged for the current task to be woken up when they can later
+make progress, such as when a network packet arrives. 
+
+By using the non-blocking Tokio I/O types, a future that performs I/O
+no longer blocks execution of other futures if the I/O they wish to
+perform cannot be performed immediately. Instead, it simply returns
+`NotReady`, and relies on a task notification to cause `poll` to be
```
rylev: Is it to early to specifically call this "task notification" or should this be more hand-wavy?

carllerche: The previous section (working with futures) will have talked about this (well, once that is written).

content/docs/io/overview.md
carllerche: Yeah, we probably should not use `promise` here given that it is overloaded.

content/docs/io/reading_writing_data.md
```diff
@@ -6,7 +6,156 @@ menu:
...
+In [the overview] we mentioned briefly that Tokio's I/O types implement
+non-blocking variants of `std::io::Read` and `std::io::Write` called
+[`AsyncRead`] and [`AsyncWrite`]. These are an integral part of Tokio's
+I/O story, and are important to understand when working with I/O code.
+
+> Note: in this section, we'll primarily talk about `AsyncRead`, but
+> `AsyncWrite` is pretty much exactly the same, just for writing data to
+> an I/O resource (like a TCP socket) instead of reading from it.
+
+So, let's take a look at [`AsyncRead`] and see what all the fuzz is
```
rylev: I believe this idiom is "what all the fuss is about", but maybe this is a variant I'm not familiar with... :-D

content/docs/io/reading_writing_data.md
```diff
@@ -6,7 +6,156 @@ menu:
...
+use std::io::Read;
+pub trait AsyncRead: Read {
+    // ...
+    // various provided methods
+    // ...
+}
+```
+
+Huh. What's going on here? Well, `AsyncRead` is really just `Read` from
+`std::io`, along with a _contract_. The documentation for `AsyncRead`
```
rylev: Should it be "additional contract" instead of just contract?

content/docs/io/reading_writing_data.md
```diff
@@ -6,7 +6,156 @@ menu:
...
+fn split(self) -> (ReadHalf<Self>, WriteHalf<Self>)
+  where Self: AsyncWrite { ... }
+```
+
+The returned `ReadHalf` implements `AsyncRead`, and the `WriteHalf`
+implements `AsyncWrite`. And crucially, we now have two *separate*
+pointers into our type, which we can pass around separately. This comes
+in handy for [`copy`], but it also means that we can pass each half to
+a different future, and handle the reads and writes completely
+independently! Behind the scenes, [`split`] uses a [`BiLock`] to ensure
```
rylev: While interesting, I wonder if the `BiLock` comment is just distracting?

rylev:
Awesome stuff! Mostly just minor nitpicks. This seems good enough to merge once the issues are addressed IMO.

jonhoo:
I'd like to write the remaining sections for I/O too before merging, because later text may end up changing earlier text :)

rylev:
@jonhoo sounds like a plan to me! 👍

AndyGauge:
👍

jonhoo:
Next section ("Using AsyncRead and AsyncWrite directly") is now ready for review :)
Next up: "Implementing Async Read/Write"

jonhoo:
Okay, by @carllerche's request, let's do a review on the sections that are here now, and then I'll continue onto the more implementation-oriented sections of I/O after this is merged :)
@rylev could you take a look at the new section + new code examples?
@AndyGauge would you have time to take a look through as well?

carllerche:
I did a quick skim and I like the direction 👍  I don't have time to do an in depth review right now, so if ya'll are good with this, ship it 😄

content/docs/io/async_read_write.md
```diff
@@ -1,12 +1,311 @@
...
+            }
+            ReadExact::Empty => panic!("poll a ReadExact after it's done"),
+        }
+
+        // We need to return the reader and the buffer, which we can only
+        // do by moving them out of self. We do this by swapping our state
+        // with an "empty" one. This _should_ be fine, because poll() requires
+        // callers to not call poll() again after Ready has been returned,
+        // so we should only ever see ReadExact::Reading when poll() is called.
+        match mem::replace(self, ReadExact::Empty) {
```
rylev: If we use an `Option<Reading>` instead of `Reading` be a variant of some `ReadExact` enum, we can avoid having to do the `mem::replace` and just use `Option::take`. Is that something we should consider?

jonhoo: You'd need to wrap both the `buffer` and `reader` in `Option`, and then `take` them both. And you'd then also have to match on `Some` twice in `poll`. It would be possible, but not sure it's cleaner (except for avoiding the `mem::replace`). The reason I chose to do it this way was primarily because it matches the real implementation in `tokio`, but I could be convinced to change it :)

content/docs/io/async_read_write.md
```diff
@@ -1,12 +1,311 @@
...
+the underlying I/O protocol requires. For TCP connections, for example,
+this usually entails closing the writing side of the TCP channel so that
+the other end receives and end-of-file in the form of a read that
+returns 0 bytes. You can often think of `shutdown` as what you _would_
+have done synchronously in the implementation of `Drop`; it's just that
+in the asynchronous world, you can't easily do something in `Drop`
+because you need to have an executor that keeps polling your writer!
+
+<!--
+Socket is safely dropped after shutdown returns Ok(Ready).
```
rylev: Should this be deleted?

jonhoo: It won't actually be visible in the rendered article since it's an HTML comment. I left it in there because it might be something we want to cover in later drafts.

rylev: You might want to mark it as "Something we might want to cover in the future" - I'm not convinced future authors will know how to interpret the comment.

jonhoo: Ah, yes, good point!

rylev:
This looks great! Awesome work!

content/docs/io/async_read_write.md
jonhoo: I misunderstood you. Made the change in e06d56e07f4e0b76b34ca1a910c6453b8f978796!

jonhoo:
Merging this now -- we can always make more changes later. @carllerche do I need to do anything special to make the changes appear on tokio.rs?

jonhoo:
nvm, it's now public! :tada:

carllerche:
automated deploys ftw!

PR #266, opened 2018-12-09
Update getting_asynchronous.md

mathijshenquet:
I was reading the documentation yesterday and I thought this particular example was a bit hard to grok at first. I think it would be more clear to me if rewritten this way.

### DIFF
--- a/content/docs/futures/getting_asynchronous.md
+++ b/content/docs/futures/getting_asynchronous.md
@@ -109,9 +109,11 @@ impl Future for HelloWorld {
         use self::HelloWorld::*;
 
         loop {
-            let socket = match *self {
+            match *self {
                 Connecting(ref mut f) => {
-                    try_ready!(f.poll())
+                    let socket = try_ready!(f.poll());
+                    let data = Cursor::new(Bytes::from_static(b"hello world"));
+                    *self = Connected(socket, data);
                 }
                 Connected(ref mut socket, ref mut data) => {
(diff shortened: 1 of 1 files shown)


carllerche:
I agree that it is confusing, unfortunately, the confusing bits are required to make the borrow checker happy.

That said, it could be that non lexical lifetimes fixes this. I'm not sure how to check.

mathijshenquet:
I just checked: on rust 1.31.0 the above compiles.

carllerche:
@mathijshenquet That makes sense as I believe that NLL is only enabled for the 2018 edition. Now, the question is how best to proceed. Should the docs use the 2018 edition for examples?

carllerche:
I think the consensus is that the website should switch to use the 2018 edition. That probably should go in a separate PR to update the Ci script and any example that breaks. Any interest in taking a stab?

carllerche:
Could you merge master? That should hopefully get CI to pass.

ultrasaurus:
@mathijshenquet it looks like this didn't get update after Tokio moved to Rust 2018, and now more time has passed and this content would need to be changed once again (I think) for the changes in futures with Rust 1.39.0. In any case4, this page has been removed from the nav while we figure out how/whether to update the old content.  So, I'm going to close this PR.

If you want to do more work on this or tokio docs in general, please join us in the tokio-docs discord channel: https://discord.gg/tokio

PR #283, opened 2019-02-02
Update netcat command in hello-world.md

mziter:
Existing command to run example in netcat did not work for me. I had to modify command to that which is in proposed change.


mziter:
Seems like issue is with version of netcat used. Will close and discuss more before proposing change.

PR #284, opened 2019-02-02
Update hello-world.md

mziter:
Modified Netcat example to be GNU specific netcat command and added quote area with information about handling if user is using a different version of netcat. Added link to NMap.org version of netcat which is newer and has downloads for all major OS's, in case user needs to install.

### DIFF
--- a/content/docs/getting-started/hello-world.md
+++ b/content/docs/getting-started/hello-world.md
@@ -204,8 +204,11 @@ You can find the full example [here][full-code].
 command starts a listening TCP socket on the previously specified port.
 
 ```bash
-$ nc -l 6142
+$ nc -l -p 6142
 ```
+> The command above is used with the GNU version of netcat that comes stock on many
+> unix based operating systems. The following command can be used with the
+> [NMap.org][NMap.org] version: `$ ncat -l -p 6142`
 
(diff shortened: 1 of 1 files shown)


content/docs/getting-started/hello-world.md
```diff
@@ -204,8 +204,11 @@ You can find the full example [here][full-code].
 command starts a listening TCP socket on the previously specified port.
 
 ```bash
-$ nc -l 6142
+$ nc -l -p 6142
 ```
+> The command above is used with the GNU version of netcat that comes stock on many
+> unix based operating systems. The following command can be used with the
+> [NMap.org][NMap.org] version: `$ ncat -l -p 6142`
```
tobz: This doesn't make sense to me.

The default example and the "nmap.org version" example as the same thing.

mziter: one is `nc ...` and one is `ncat`. If it is more confusing with the same arguments then maybe something like "if using nmap org version, substitute ncat for nc" or something like that?

tobz: Ah, gotcha.  I was focused on the arguments, not the binary name. :)

PR #295, opened 2019-03-18
Update examples to 2018 edition

bIgBV:
I'm updating the examples as I go through them. To test weather or not this works, I ran the examples to test and ensured that everything ran as expected and no warnings were generated by the compiler.

### DIFF
--- a/content/docs/futures/basic.md
+++ b/content/docs/futures/basic.md
@@ -33,9 +33,6 @@ trait Future {
 Let's implement it for our "hello world" future:
 
 ```rust
-# #![deny(deprecated)]
-extern crate futures;
-
 // `Poll` is a type alias for `Result<Async<T>, E>`
 use futures::{Future, Async, Poll};
 
@@ -85,9 +82,6 @@ In our case, let's print the future to STDOUT. We will do that by implementing a
 `Display` future.
 
 ```rust
-# #![deny(deprecated)]
-extern crate futures;
-
 use futures::{Future, Async, Poll};
 use std::fmt;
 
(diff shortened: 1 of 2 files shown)


bIgBV:
I ran the ci `test.sh` script, but I'm not entirely sure what's causing the errors, since the examples work standalone.

carllerche:
Does this command (in the ci script) have to be updated?

https://github.com/tokio-rs/website/blob/6a012bacfb88574a7be96f9f8e1261f8540843b8/ci/test.sh#L17

carllerche:
PS, thanks for updating!

bIgBV:
I'm not sure, if I understand the line correctly, it takes the passed in markdown file and then runs the rustc on it with the `test` flag, right? If that's the case it should work, but something seems to be missing.

carllerche:
@bIgBV If `rust-skeptic` works, we can try switching to it!

carllerche:
My guess is that the command does need to specify the crates to "extern". Without the explicit `extern` lines, how would it know which crates to link?

bIgBV:
Yeah, that makes sense. I'll try to get `rust-skeptic` to work and hopefully, it can handle both the situations while the codebase is in transition. I think it should because I'm guessing it just somehow invoked `cargo-test` internally.

ultrasaurus:
@bIgBV time has passed and now this content would need to be changed once again some of this has moved into the Rust language.  For now this page has been removed from the nav and I'm going to close this PR.

If you want to do more work on this or tokio docs in general, please join us in the tokio-docs discord channel: https://discord.gg/tokio

PR #301, opened 2019-04-08
Proposal for minor improvement/clarification to the overview of futures

jonasbn:
The single bullet in the overview kind of stand out as being very _technical_ compared to the others and it does not clarify the subtleties of the `()`. As a *Rust* _n00b_ I found this confusing, since the abstraction level in the documentation is not consistent.


carllerche:
Thanks!

PR #304, opened 2019-04-22
Improvement/clarification to points related to overview of futures

jonasbn:
In PR #301, I proposed a minor clarification to the point on timeouts

Since the same listing of points is reused, I propose applying the same
clarification as outlined in PR #301.

The same listing is used at:

- https://tokio.rs/docs/futures/overview/ (PR #301)
- https://tokio.rs/docs/getting-started/futures/

Explanation lifted from #301:

> The single bullet in the overview kind of stand out as being very
> _technical_ compared to the others and it does not clarify the
> subtleties of the `()`. As a *Rust* _n00b_ I found this confusing,
> since the abstraction level in the documentation is not consistent.


carllerche:
Thanks! Much appreciated

PR #310, opened 2019-05-05
Make futures crate usage consistent in echo server

polybuildr:
The `futures` crate isn't used in any of the visible portion on the tokio.rs
site, so the extern crate declaration should be commented out from the code
samples.

Also, since the crate is used in the commented out portion of the code, it
should also be added to Cargo.toml, but commented out.


polybuildr:
Ping?

PR #313, opened 2019-05-14
Rewrite comment in spawning docs

roignpar:
In the background processing code example there seem to be 2 misplaced comments.
I removed the one about writing to the socket since it doesn't make sense in this context and moved the drop comment right before `.map(|_| ())` where it should be, as far as I can tell.
Let me know if the first comment should be replaced instead.

### DIFF
--- a/content/docs/futures/spawning.md
+++ b/content/docs/futures/spawning.md
@@ -300,14 +300,12 @@ tokio::run(lazy(|| {
             // Each spawned task will have a clone of the sender handle.
             let tx = tx.clone();
 
-            // In this example, "hello world" will be written to the
-            // socket followed by the socket being closed.
             io::read_to_end(socket, vec![])
-                // Drop the socket
                 .and_then(move |(_, buf)| {
                     tx.send(buf.len())
                         .map_err(|_| io::ErrorKind::Other.into())
                 })
+                // Drop the socket
                 .map(|_| ())
                 // Write any error to STDOUT
                 .map_err(|e| println!("socket error = {:?}", e))
--- a/content/docs/futures/spawning.md
+++ b/content/docs/futures/spawning.md
@@ -300,6 +300,8 @@ tokio::run(lazy(|| {
             // Each spawned task will have a clone of the sender handle.
             let tx = tx.clone();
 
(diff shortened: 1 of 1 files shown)


roignpar:
Looking at it again, I believe that there is no socket to drop.
Should the drop comment also be removed? Or changed to `Drop tx handle`?

content/docs/futures/spawning.md
```diff
@@ -300,14 +300,12 @@ tokio::run(lazy(|| {
             // Each spawned task will have a clone of the sender handle.
             let tx = tx.clone();
 
-            // In this example, "hello world" will be written to the
-            // socket followed by the socket being closed.
```
carllerche: Instead of removing the comment entirely, should we update it to describe what the code snippet is doing?

content/docs/futures/spawning.md
```diff
@@ -300,14 +300,12 @@ tokio::run(lazy(|| {
...
             io::read_to_end(socket, vec![])
-                // Drop the socket
```
carllerche: As of mentioned, the drop happens here. How would you feel about explicitly calling the drop fn to make it clearer?

roignpar: It would indeed be clearer.
On the other hand, it's not something one would need to write in real use cases and none of the other examples use `drop`.

The reason I moved the comment in the first place was that I noticed this piece of code was probably copied from the previous "Processing inbound sockets" code example which has the same comment right before `.map(|_| ())`. It was just a confusion.
I think it's best to just leave the comment as it was.

content/docs/futures/spawning.md
roignpar: Something along the lines of `In this example, all bytes read from the socket are placed into a Vec`?

carllerche: That sounds good.

content/docs/futures/spawning.md
carllerche: Ok :+1:

roignpar:
@carllerche thank you for taking the time to review and all your contributions to the rust ecosystem!

PR #332, opened 2019-09-04
Fix netcat command

MarkKoester:
I just tested this command and I don't think you use the -p flag with the -l flag. The correct usage is to just specify the port after the -l flag.


carllerche:
I'm pretty sure `netcat` varies per platform, so we really need to stop using it (i have a recollection of changing the netcat command a few times already).

patrick0210:
The command `$ nc -l 6142` does not work with the GNU netcat (version 0.7.1). The netcat of nmap.org accepts both (I have built the latest version). I use Linux.

PR #347, opened 2019-11-12
Update hello world for async-await

vorot93:
Let's get this transition going!


vorot93:
Closing in favor of #352.

PR #352, opened 2019-11-15
Update the tutorial to Tokio 0.2

vorot93:
This is a combined PR that will include the community effort to upgrade existing tutorial to Tokio 0.2 async/await.

### DIFF
--- a/content/docs/getting-started/futures.md
+++ b/content/docs/getting-started/futures.md
@@ -6,9 +6,9 @@ menu:
     parent: getting_started
 ---
 
-Let's take a closer look at futures. Tokio is built on top of the [`futures`] crate
-and uses its runtime model. This allows Tokio to interop with other libraries also
-using the [`futures`] crate.
+Let's take a closer look at futures. Tokio is built on top of `task` and `future` modules
+found in the Rust standard library as well as the `futures` crate.
+This allows Tokio to interop with other libraries also built on the same foundation.
 
 **Note**: This runtime model is very different than async libraries found in
 other languages. While, at a high level, APIs can look similar, the way code
@@ -21,7 +21,7 @@ default and see how this differs from Tokio's asynchronous model.
 
 # Synchronous Model
 
-First, let's talk briefly about the synchronous (or blocking) model that the
+First, let's talk briefly about the synchronous (or blocking) model that most of the
 Rust [standard library] uses.
 
 ```rust
@@ -117,7 +117,7 @@ might pass the future to a function.
 
 ```rust,ignore
 let response_is_ok = response_future
-    .map(|response| {
+    .map_ok(|response| {
         response.status().is_ok()
     });
 
@@ -163,27 +163,22 @@ The `Future` trait is as follows:
 ```rust,ignore
 trait Future {
     /// The type of the value returned when the future completes.
-    type Item;
-
-    /// The type representing errors that occurred while processing the computation.
-    type Error;
+    type Output;
 
     /// The function that will be repeatedly called to see if the future
     /// has completed or not
-    fn poll(&mut self) -> Result<Async<Self::Item>, Self::Error>;
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output>;
 }
 ```
 
-For now it's just important to know that futures have two associated types: `Item`
-and `Error`. `Item` is the type of the value that the `Future` will yield when it
-completes. `Error` is the type of Error that the `Future` may yield if there's an
-error before that prevents the `Future` from completing successfully.
+For now it's just important to know that futures have an associated type: `Output`.
+It is the type of the value that the `Future` will yield when it completes.
 
-Finally, `Future`s have one method named `poll`. We won't go into too much detail
+Additionally, `Future`s have one method named `poll`. We won't go into too much detail
 about `poll` in this section since you don't need to know about `poll` to use
 futures with combinators. The only thing to be aware for now is that `poll` is
 what the runtime will call in order to see if the `Future` is complete yet or not.
-If you're curious: `Async` is an enum with values `Ready(Item)` or `NotReady` which
+If you're curious: `Poll` is an enum with values `Ready(Item)` or `Pending` which
 informs the runtime of if the future is complete or not.
 
 In a future section, we'll be implementing a `Future` from scratch including writing
@@ -215,12 +210,9 @@ trait Stream {
     /// The type of the value yielded by the stream.
     type Item;
 
-    /// The type representing errors that occurred while processing the computation.
-    type Error;
-
     /// The function that will be repeatedly called to see if the stream has
     /// another value it can yield
-    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error>;
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>>
 }
 ```
 
--- a/content/docs/getting-started/futures.md
+++ b/content/docs/getting-started/futures.md
@@ -6,8 +6,8 @@ menu:
     parent: getting_started
 ---
 
-Let's take a closer look at futures. Tokio is built on top of `task` and `future` modules
-found in the Rust standard library as well as the `futures` crate.
+Let's take a closer look at futures. Tokio is built on top of [`task`] and [`future`] modules
+found in the Rust standard library, as well as on the [`futures`] crate.
 This allows Tokio to interop with other libraries also built on the same foundation.
 
 **Note**: This runtime model is very different than async libraries found in
@@ -172,13 +172,13 @@ trait Future {
 ```
 
 For now it's just important to know that futures have an associated type: `Output`.
-It is the type of the value that the `Future` will yield when it completes.
+This is the type of the value that the `Future` will yield when it completes.
 
-Additionally, `Future`s have one method named `poll`. We won't go into too much detail
+Additionally, `Future`s have one method, named `poll`. We won't go into too much detail
 about `poll` in this section since you don't need to know about `poll` to use
-futures with combinators. The only thing to be aware for now is that `poll` is
+futures with async/await. The only thing to be aware for now is that `poll` is
 what the runtime will call in order to see if the `Future` is complete yet or not.
-If you're curious: `Poll` is an enum with values `Ready(Item)` or `Pending` which
+If you're curious: [`Poll`] is an enum with values `Ready(Item)` or `Pending` which
 informs the runtime of if the future is complete or not.
 
 In a future section, we'll be implementing a `Future` from scratch including writing
@@ -219,9 +219,12 @@ trait Stream {
 Streams come with their own set of combinators and will be covered in more depth
 in the [working with futures][working-with-streams] section.
 
+[`task`]: https://doc.rust-lang.org/std/task
+[`future`]: https://doc.rust-lang.org/std/future
 [`futures`]: {{< api-url "futures" >}}
 [standard library]: https://doc.rust-lang.org/std/
 [c10k]: https://en.wikipedia.org/wiki/C10k_problem
+[`Poll`]: https://doc.rust-lang.org/std/task/enum.Poll.html
 [`TcpStream`]: {{< api-url "tokio" >}}/net/struct.TcpStream.html
 [`ErrorKind::WouldBlock`]: https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.WouldBlock
 [working-with-streams]: {{< ref "/docs/futures/streams.md" >}}
--- a/content/docs/getting-started/hello-world.md
+++ b/content/docs/getting-started/hello-world.md
@@ -29,23 +29,11 @@ Next, add the necessary dependencies in `Cargo.toml`:
 
 ```toml
 [dependencies]
-tokio = "0.1"
+tokio = "0.2.0-alpha"
 ```
 
-and the crates and types into scope in `main.rs`:
-
-```rust
-# #![deny(deprecated)]
-extern crate tokio;
-
-use tokio::io;
-use tokio::net::TcpStream;
-use tokio::prelude::*;
-# fn main() {}
-```
-
-Here we use Tokio's own [`io`] and [`net`] modules. These modules provide the same
-abstractions over networking and I/O-operations as the corresponding modules in `std`
+We are going to use Tokio's own [`net`] module. This module provides the same
+abstractions over networking and I/O-operations as the corresponding module in `std`
 with a small difference: all actions are performed asynchronously.
 
 # Creating the stream
@@ -54,14 +42,11 @@ The first step is to create the `TcpStream`. We use the `TcpStream` implementati
 provided by Tokio.
 
 ```rust
-# #![deny(deprecated)]
-# extern crate tokio;
-#
-# use tokio::net::TcpStream;
+use tokio::net::TcpStream;
+
 fn main() {
-    // Parse the address of whatever server we're talking to
-    let addr = "127.0.0.1:6142".parse().unwrap();
-    let client = TcpStream::connect(&addr);
+    let addr = "127.0.0.1:6142";
+    let client = TcpStream::connect(addr);
 
     // Following snippets come here...
 }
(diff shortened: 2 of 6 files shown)


content/docs/getting-started/futures.md
```diff
@@ -6,9 +6,9 @@ menu:
     parent: getting_started
 ---
 
-Let's take a closer look at futures. Tokio is built on top of the [`futures`] crate
-and uses its runtime model. This allows Tokio to interop with other libraries also
-using the [`futures`] crate.
+Let's take a closer look at futures. Tokio is built on top of `task` and `future` modules
```
hawkw: Nit: should `task` and `future` be linked? `futures` was linked previously.

content/docs/getting-started/futures.md
```diff
@@ -6,9 +6,9 @@ menu:
...
+found in the Rust standard library as well as the `futures` crate.
```
hawkw: Nit: probably should link to `futures` here.

hawkw:
```suggestion
found in the Rust standard library, as well as on the `futures` crate.
```

content/docs/getting-started/futures.md
```diff
@@ -163,27 +163,22 @@ The `Future` trait is as follows:
...
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output>;
 }
 ```
 
-For now it's just important to know that futures have two associated types: `Item`
-and `Error`. `Item` is the type of the value that the `Future` will yield when it
-completes. `Error` is the type of Error that the `Future` may yield if there's an
-error before that prevents the `Future` from completing successfully.
+For now it's just important to know that futures have an associated type: `Output`.
+It is the type of the value that the `Future` will yield when it completes.
```
hawkw:
```suggestion
This is the type of the value that the `Future` will yield when it completes.
```

content/docs/getting-started/futures.md
```diff
@@ -163,27 +163,22 @@ The `Future` trait is as follows:
...
 
-Finally, `Future`s have one method named `poll`. We won't go into too much detail
+Additionally, `Future`s have one method named `poll`. We won't go into too much detail
 about `poll` in this section since you don't need to know about `poll` to use
 futures with combinators. The only thing to be aware for now is that `poll` is
 what the runtime will call in order to see if the `Future` is complete yet or not.
-If you're curious: `Async` is an enum with values `Ready(Item)` or `NotReady` which
+If you're curious: `Poll` is an enum with values `Ready(Item)` or `Pending` which
```
hawkw: Nit: should this link to https://doc.rust-lang.org/std/task/enum.Poll.html?

content/docs/getting-started/futures.md
```diff
@@ -163,27 +163,22 @@ The `Future` trait is as follows:
...
-and `Error`. `Item` is the type of the value that the `Future` will yield when it
-completes. `Error` is the type of Error that the `Future` may yield if there's an
-error before that prevents the `Future` from completing successfully.
+For now it's just important to know that futures have an associated type: `Output`.
+It is the type of the value that the `Future` will yield when it completes.
 
-Finally, `Future`s have one method named `poll`. We won't go into too much detail
+Additionally, `Future`s have one method named `poll`. We won't go into too much detail
 about `poll` in this section since you don't need to know about `poll` to use
 futures with combinators. The only thing to be aware for now is that `poll` is
```
hawkw: I might change "combinators" to "async/await" here:
```suggestion
futures with async/await. The only thing to be aware for now is that `poll` is
```

content/docs/getting-started/futures.md
```diff
@@ -163,27 +163,22 @@ The `Future` trait is as follows:
...
 
-For now it's just important to know that futures have two associated types: `Item`
-and `Error`. `Item` is the type of the value that the `Future` will yield when it
-completes. `Error` is the type of Error that the `Future` may yield if there's an
-error before that prevents the `Future` from completing successfully.
+For now it's just important to know that futures have an associated type: `Output`.
+It is the type of the value that the `Future` will yield when it completes.
 
-Finally, `Future`s have one method named `poll`. We won't go into too much detail
+Additionally, `Future`s have one method named `poll`. We won't go into too much detail
```
hawkw: Nit: I think there should be a comma here
```suggestion
Additionally, `Future`s have one method, named `poll`. We won't go into too much detail
```

content/docs/getting-started/futures.md
```diff
@@ -163,27 +163,22 @@ The `Future` trait is as follows:
...
 about `poll` in this section since you don't need to know about `poll` to use
 futures with combinators. The only thing to be aware for now is that `poll` is
 what the runtime will call in order to see if the `Future` is complete yet or not.
```
hawkw: A comparison I like to make that may be worth including is that the relationship between `Future::poll` and the `async`/`await` keywords is very similar to the relationship between `Iterator::next` and `for` loops: you don't need to be aware of the details of how `next` is implemented to iterate over a slice in a `for` loop. Futures and async/await are similar.

content/docs/getting-started/futures.md
```diff
@@ -215,12 +210,9 @@ trait Stream {
     /// The type of the value yielded by the stream.
     type Item;
 
-    /// The type representing errors that occurred while processing the computation.
-    type Error;
-
     /// The function that will be repeatedly called to see if the stream has
     /// another value it can yield
-    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error>;
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>>
```
hawkw: Nit: should this be 
```suggestion
    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>>
```
since the Rust 2018 idiom lints suggests that? 

I notice that the standard library definition of `Future::poll` does not use this syntax though, so maybe this should be the same?

content/docs/getting-started/hello-world.md
```diff
@@ -71,93 +56,74 @@ Next, we'll add some to the `client` `TcpStream`. This asynchronous task now cre
...
+
+                // Process stream here.
+            }
+            Err(err) => {
+                // In our example, we are only going to log the error to STDOUT.
+                println!("connection error = {:?}", err);
+            }
+        };
+    };
+}
```
hawkw: In this example, the future is never run. I think that, in the previous example, that is okay, since we hid the `fn main()` and it just looked like code somewhere inside a program. Here, though, we are showing `fn main()` at the top of the example. IMO, this gives the incorrect implication that users can write code like this and expect it to work.

Maybe we should hide the `fn main()` in this example?

content/docs/getting-started/hello-world.md
```diff
@@ -71,93 +56,74 @@ Next, we'll add some to the `client` `TcpStream`. This asynchronous task now cre
...
+
+    let future = async move {
+        match client.await {
+            Ok(_stream) => {
+                println!("created stream");
+
+                // Process stream here.
+            }
+            Err(err) => {
+                // In our example, we are only going to log the error to STDOUT.
```
hawkw: Maybe this should have a comment saying that the `TcpStream::connect` future has an `Output` type of `io::Result<TcpStream>`? Previously, there was a comment about `Future::Error`, but since those no longer exist, we should explain where the error is coming from here.

content/docs/getting-started/hello-world.md
```diff
@@ -71,93 +56,74 @@ Next, we'll add some to the `client` `TcpStream`. This asynchronous task now cre
...
+                println!("connection error = {:?}", err);
+            }
+        };
+    };
+}
 ```
 
-The call to `TcpStream::connect` returns a [`Future`] of the created TCP stream.
+Inside the block the call to `TcpStream::connect` returns a [`Future`] of the created TCP stream.
```
hawkw: The call to `TcipStream::connect` is not inside an `async` block here?

```suggestion
The call to `TcpStream::connect` returns a [`Future`] of the created TCP stream.
```

content/docs/getting-started/hello-world.md
```diff
@@ -71,93 +56,74 @@ Next, we'll add some to the `client` `TcpStream`. This asynchronous task now cre
...
-sequenced. A `Future` is a value representing some computation that will complete at
-some point in the future.
+with a value representing the work of creating a TCP stream.
 
-It's worth reiterating that returned futures are lazy, i.e., no work is performed when
-calling the combinator. Instead, once all the asynchronous steps are sequenced, the
-final `Future` (representing the entire task) is 'spawned' (i.e., run). This is when
-the previously defined work starts getting run. In other words, the code we've written
-so far does not actually create a TCP stream.
+Note the `async`-block that follows. Contents of the async block are not executed immediately but are rather
```
hawkw: Some minor nits:
```suggestion
Note the `async` block that follows. Contents of the `async` block are not executed immediately, but are rather
```

content/docs/getting-started/hello-world.md
```diff
@@ -71,93 +56,74 @@ Next, we'll add some to the `client` `TcpStream`. This asynchronous task now cre
...
+    };
+}
 ```
 
-The call to `TcpStream::connect` returns a [`Future`] of the created TCP stream.
+Inside the block the call to `TcpStream::connect` returns a [`Future`] of the created TCP stream.
 We'll learn more about [`Futures`] later in the guide, but for now you can think of
 a [`Future`] as a value that represents something that will eventually happen in the
 future (in this case the stream will be created). This means that `TcpStream::connect` does
 not wait for the stream to be created before it returns. Rather it returns immediately
```
hawkw: Also, I know this was already here, but I think there should be a comma after "Rather"
```suggestion
not wait for the stream to be created before it returns. Rather, it returns immediately
```

content/docs/getting-started/hello-world.md
hawkw: Also, we use "rather" in the last sentence of the previous paragraph. What about something like:
```suggestion
Note the `async` block that follows. Like the `TcpStream::connect` future, the contents of the `async` block are not executed immediately. Instead, they are
```

content/docs/getting-started/hello-world.md
```diff
@@ -71,93 +56,74 @@ Next, we'll add some to the `client` `TcpStream`. This asynchronous task now cre
...
-some point in the future.
+with a value representing the work of creating a TCP stream.
 
-It's worth reiterating that returned futures are lazy, i.e., no work is performed when
-calling the combinator. Instead, once all the asynchronous steps are sequenced, the
-final `Future` (representing the entire task) is 'spawned' (i.e., run). This is when
-the previously defined work starts getting run. In other words, the code we've written
-so far does not actually create a TCP stream.
+Note the `async`-block that follows. Contents of the async block are not executed immediately but are rather
+stored in an object whose result can be computed later. Inside the `async` block we can `await` on the
```
hawkw:
```suggestion
stored in an object whose result can be computed later. Inside the `async` block, we can `await` the
```

content/docs/getting-started/hello-world.md
```diff
@@ -71,93 +56,74 @@ Next, we'll add some to the `client` `TcpStream`. This asynchronous task now cre
...
+created future so that any code that follows is executed with the result of the `await`-ed [`Future`]'s computation.
```
hawkw:
```suggestion
created future, so that any code that follows is executed with the result of the `await`-ed [`Future`]'s computation.
```

content/docs/getting-started/hello-world.md
```diff
@@ -71,93 +56,74 @@ Next, we'll add some to the `client` `TcpStream`. This asynchronous task now cre
...
-we acknowledge errors.
-
 Next, we will process the stream.
 
 # Writing data
 
 Our goal is to write `"hello world\n"` to the stream.
 
-Going back to the `TcpStream::connect(addr).and_then` block:
+Going back to the `TcpStream::connect(addr)` block:
```
hawkw: `TcpStream::connect(addr)` isn't in a block, how about:
```suggestion
Going back to the `async` block that `await`s the `TcpStream::connect` future:
```
or just
```suggestion
Going back to the `async` block in the previous example:
```

content/docs/getting-started/hello-world.md
```diff
@@ -71,93 +56,74 @@ Next, we'll add some to the `client` `TcpStream`. This asynchronous task now cre
...
+    let client = TcpStream::connect(addr);
+
+    let future = async move {
+        if let Ok(mut stream) = client.await {
+            println!("created stream");
+
+            let result = stream.write_all(b"hello world\n").await;
+
+            println!("wrote to stream; success={:?}", result.is_ok());
+        }
```
hawkw: Before we added the call to `write_all`, this code used a `match`, but now it uses an `if let`...why not continue using the match? That way, we still show error handling, and less code changes between the two examples.

content/docs/getting-started/hello-world.md
```diff
@@ -71,93 +56,74 @@ Next, we'll add some to the `client` `TcpStream`. This asynchronous task now cre
...
+            Ok(_stream) => {
+                println!("created stream");
+
+                // Process stream here.
+            }
+            Err(err) => {
+                // In our example, we are only going to log the error to STDOUT.
+                println!("connection error = {:?}", err);
+            }
+        };
```
hawkw: We might want to consider modifying this code to just use the `?` operator — removing error handling boilerplate would let users focus on the new content being introduced. Maybe something like:
```rust
let future = async move {
    let stream = client.await?;
    println!("created stream");
# drop(stream); // this lets the `stream` local appear to be used so we don't need `_stream`

    // process stream here

    Ok(())
};
```
and add a comment explaining that we can use `?` in an `async` block the way we can in other code.

content/docs/getting-started/hello-world.md
```diff
@@ -175,27 +141,61 @@ different pros and cons. In this example, we will use the default executor of th
...
+            let result = stream.write_all(b"hello world\n").await;
+
+            println!("wrote to stream; success={:?}", result.is_ok());
+        }
+    };
+
+    let rt = Runtime::new().unwrap();
+    println!("About to create the stream and write to it...");
+    rt.block_on(future);
+    println!("Stream has been created and written to.");
```
hawkw: Since these are "getting started" docs, maybe we should just use the batteries-included `#[tokio::main]` attribute, rather than creating a `Runtime` manually (more of an advanced API). I notice that the previous version used `tokio::run`, which, IMO, is more or less the tokio 0.1 equivalent of `#[tokio::main]`, so perhaps we should continue using the simpler API?

content/docs/getting-started/hello-world.md
```diff
@@ -175,27 +141,61 @@ different pros and cons. In this example, we will use the default executor of th
...
+
+    if let Ok(mut stream) = client.await {
+        println!("created stream");
+
+        let result = stream.write_all(b"hello world\n").await;
+
+        println!("wrote to stream; success={:?}", result.is_ok());
+    }
+}
+```
```
hawkw: Ah, I see we introduce `#[tokio::main]` here...I still think it _might_ be better to show the easier API first, and then say "okay, here is what `tokio::main` actually _does_" and show code that manually creates a runtime.

content/docs/getting-started/runtime.md
```diff
@@ -54,29 +54,25 @@ trait directly.
...
-});
+    tokio::spawn(async move {
+      let my_inner_value = my_inner_future.await;
+      println!("Got a value {:?} from second future", my_inner_value);
+    });
+  }
+};
 
-tokio::run(program);
+tokio::runtime::Runtime::new().unwrap().block_on(program);
```
hawkw: why not just use `#[tokio::main]` in the example?

vorot93: We would have to define `my_outer_stream` which I wanted to avoid.

hawkw: ah, I see, that makes sense. we could always show a function like
```rust
fn make_stream() -> impl Stream<Item = usize> + Send + Sync {
# tokio::future::ready(1)
    // ...
}
```
and then call that function in the example. although I am not sure if this is any better.

vorot93:
Closing in favour of #357

PR #354, opened 2019-11-19
Add init support for mdbook

kpp:


kpp:
Oops. Wrong target base.

PR #355, opened 2019-11-21
idea: 0.2 version branch with different header color

ultrasaurus:
I was taking a look at https://github.com/tokio-rs/website/pull/352 and 
looking at where to add https://github.com/tokio-rs/doc-push/issues/98

I think it will help working on 0.2 docs to have a in-progress version posted somewhere
that is visually different from the current / full-tested docs.  

This is a suggestion to have a different header color in a branch and collect 0.2 changes together

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -29,6 +29,8 @@ menu:
 
 
 params:
+  version: "0.2.0-alpha.6"
+  status: "alpha"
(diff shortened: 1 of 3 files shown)


ultrasaurus:
https://github.com/tokio-rs/website/pull/396 was merged with nicer visual treatment for this idea

PR #357, opened 2019-11-26
start superficial 0.2.x update of docs

carllerche:


content/docs/getting-started/futures.md
```diff
@@ -161,75 +127,23 @@ We'll take a closer look at this poll based model in the next section.
...
+pub trait Future {
+    /// The type of value produced on completion.
+    type Output;
 
-    /// The function that will be repeatedly called to see if the stream has
-    /// another value it can yield
-    fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error>;
+    /// Attempt to resolve the future to a final value, registering
+    /// the current task for wakeup if the value is not yet available.
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output>;
```
LucioFranco:
```suggestion
    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output>;
```

content/docs/getting-started/runtime.md
```diff
@@ -6,90 +6,63 @@ menu:
...
 ```
 
-Again spawning tasks can happen within other futures or streams allowing multiple
-things to happen concurrently. In the above example we're spawning the inner future
-from within the outer stream. Each time we get a value from the stream we'll simply
-run inner future.
+Again spawning tasks can happen within other futures or streams allowing
+multiple things to happen concurrently. In the above example we're spawning the
+inner future from within the outer stream. Each time we get a value from the
+stream we'll simply run inner future.
```
jkczyz: Runtime code explanation is outdated. It refers to an older example.

PR #379, opened 2019-12-16
Correct grammar

kellytk:


carllerche:
looks like there is a conflict :(

PR #392, opened 2020-01-08
Update Example link to point to master.

scode:
The current link is old and points to a version that predates
async/await. There is no 0.2.x branch that I can see corresponding to
the 0.1.x branch, so I'm proposing pointing to master.


ultrasaurus:
This seems like it would be high value to fix!

PR #393, opened 2020-01-09
simplify hello-world getting started

ultrasaurus:
suggestion for how to simplify "hello world" 

- app wasn't using `io` so removed from text
- adjust text to describe as a client app (since that's what it is)
- use socat which hopefully doesn't have the same cross-platform issues as we were running into with netcat


ultrasaurus:
closing this, since I've got Netlify preview hooked up on PR#394 which includes this

PR #394, opened 2020-01-09
getting started rewrite + intro to tokio features

ultrasaurus:
1. simplify "hello world"
   - use socat which hopefully doesn't have the same cross-platform issues as we were running into with netcat
   - intro network utils first (which also provides a gentle intro to networking)
   - app wasn't using `io` so removed from text
   - adjust text to describe hello-world as a client app (since that's what it is)

2. new page: intro to feature dependencies. The idea is to just walk through the few feature that are used in the "hello world" app so that people understand the concept. 
   - outline  the reasons why to use features=full or select specific features.  
   - Later we'd like to add a topic page that goes into more detail about all of the feature groupings, and this page would then link to that (or we could even make this into a smaller call out when we have that complete info). -- https://github.com/tokio-rs/website/issues/401 for feature topic page

3. Network utilities section -- this is a bit bare bones right now.  I didn't feel good about using anything in the guide without install instructions and it's a bit verbose without instructions.  -- open issue for x-platform testing and maybe using `nc` instead of `telnet` https://github.com/tokio-rs/website/issues/407


ultrasaurus:
Would appreciate if @carllerche or someone could review this for technical accuracy -- I have a couple of folks who are new(er) to tokio and will review for clarity, but want to make sure I'm not misleading them!

content/docs/getting-started/hello-features.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+or lightweight cooperative multi-tasking
+(often required for low-profile clients): 
+
+```
+tokio = { version = "0.2", features = ["macros", "rt-core"] }
+```
+
+## TCPStream connect
```
LucioFranco:
```suggestion
## TcpStream connect
```
a bit more rusty

LucioFranco: but I don't mind either way :)

content/docs/getting-started/hello-features.md
```diff
@@ -0,0 +1,92 @@
...
+1. `dns` for converting the string `127.0.0.1:6142` to an IP address -- Tokio 
+will need to do a DNS lookup just in case.
+2. `tcp` for handling the TCP connection
+
+
+## Writing to the socket
+
+Writing to the stream requires `io-util`:
+
+```
```
LucioFranco:
```suggestion
```
	use tokio::io::AsyncWriteExt;
```

may want to add this and mention that `io-util` enables this import which then enables you to use `write`.

File: content/docs/getting-started/hello-features.md [hunk 3]
```diff
@@ -0,0 +1,92 @@
...
+    let result = stream.write(b"hello world\n").await;
+```
+
+## Learn more about features
+
+The team is working to annotate the [`API reference`] to indicate which
+features are need for different APIs.  For now, you can check out the tokio
+[Cargo.toml](https://github.com/tokio-rs/tokio/blob/master/tokio/Cargo.toml)
+lists all of the features and their dependencies.
```
LucioFranco: I would maybe add something along the lines of if you see something is missing please open an issue and well answer it asap.

content/docs/getting-started/hello-features.md
```diff
@@ -0,0 +1,92 @@
...
+
+
+
+[`API reference`]: https://docs.rs/tokio/0.2.8/tokio/
```
LucioFranco:
```suggestion
[`API reference`]: https://docs.rs/tokio/
```

Might make more sense to let it resolve to the latest?

LucioFranco:
also feel free to ping me for more reviews

content/docs/getting-started/hello-features.md
ultrasaurus: I like `use tokio::prelude::*;` which I used in the simplified hello-world example (https://github.com/tokio-rs/website/pull/393/files) which then doesn't need `use tokio::io::AsyncWriteExt`

though probably would be good to mention what `io-util` enables -- I assume it is also AsyncReadExt?

content/docs/getting-started/hello-features.md
ultrasaurus: great!

content/docs/getting-started/hello-features.md
LucioFranco: This may be more of a personal thing but I very very very much dislike preludes. I think it makes the benefit of code locality go away. Basically, if I am reading code and there are no `*` imports then I can resolve any local type to either being defined in the file or `use`ed at the top. To me when reading code this is super important. So for this reason I dislike using prelude. But up to you how you want to write this part :) That's just my personal 2cents.

content/docs/getting-started/hello-features.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+## TcpStream connect
+
+As we start to build the app, we'll need more features.  The line of code
+where we call `TcpStream::connect` requires two features: 
+
+
+```
+    let stream = TcpStream::connect("127.0.0.1:6142").await.unwrap();
```
jenmei: Add `mut` (for writing later).

content/docs/getting-started/hello-features.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+
+1. `dns` for converting the string `127.0.0.1:6142` to an IP address -- Tokio
```
jenmei: Since this section is talking about using fewer features, should we mention that we can skip `dns` by using an IP address? E.g. here's a version of the code that doesn't need `dns` (and skips the prelude).

src/main.rs:
```rust
use std::net::{IpAddr, Ipv4Addr, SocketAddr};
use tokio::io::AsyncWriteExt;
use tokio::net::TcpStream;

#[tokio::main]
async fn main() {
  let addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 6142);
  let mut stream = TcpStream::connect(addr).await.unwrap();

  let result = stream.write(b"hello world\n").await;
}
```

Cargo.toml:
```toml
[dependencies]
tokio = { version = "0.2", features = ["macros", "rt-threaded", "tcp", "io-util"] }
```

content/docs/getting-started/hello-features.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Often using all features is fine 
+
+When writing an app, we'll often use all of Tokio's features to accelerate
+our development time:
+
+```toml
+[dependencies]
+tokio = { version = "0.2", features = ["full"] }
+```
```
jenmei: I suggest adding filenames before each code example. E.g.
````md
Cargo.toml:
```toml
  [dependencies]
  tokio = { version = "0.2", features = ["full"] }
```
````

File: content/docs/getting-started/hello-features.md [hunk 8]
```diff
@@ -0,0 +1,91 @@
...
+servers, we want to limit the dependencies to decrease compile time --
+not just for us, but also for the users of our crates.
+
+Let's take a look at our tiny client app and see what features are needed.
+
+## tokio main
+
+If we change our code, so that it only uses the `#[tokio::main]` macro:
+
+```
```
jenmei: Since we're essentially rewriting the hello world example, it might be nice to include the entire file (since it's just a couple of lines more).

jenmei:
This is great!

I suggest we add a section at the beginning that says what audience we're writing this for (e.g. people new to Tokio but not Rust). 

Also I would add a section saying what version this code was tested with and mention the min. version of Rust that works with this tutorial.

Also, I think it would be helpful if we preface code blocks with filenames.

content/docs/getting-started/hello-features.md
ultrasaurus: right but the whole file will have more dependencies... what I am attempting to explain is that by *only* including `#[tokio::main]` with a program that actually does nothing, I still need to include two features: "macros" and either  ("rt-threaded" or "rt-core")  

Maybe I need text that say something exactly like that.  Hmmm.

content/docs/getting-started/hello-features.md
ultrasaurus: I like this idea.

content/docs/getting-started/hello-features.md
ultrasaurus: Well for some reason mut is not needed in the hello-world app example in this PR

```
use tokio::net::TcpStream;
use tokio::prelude::*;

#[tokio::main]
async fn main() {
    let mut stream = TcpStream::connect("127.0.0.1:6142").await.unwrap();
    println!("created stream");

    let result = stream.write(b"hello world\n").await;
    println!("wrote to stream; success={:?}", result.is_ok());
}
```

I'll have to look back at other places I needed to declare `mut` in order to write. Hmmm.

content/docs/getting-started/hello-features.md
ultrasaurus: this is a great example!  I think `AsyncWriteEx`t vs `prelude` is unrelated, and it could use "rt-core" or "rt-threaded", but the key point is that "dns" can be omitted by specifying `SocketAddr`

```
use std::net::{IpAddr, Ipv4Addr, SocketAddr};
use tokio::prelude::*;
use tokio::net::TcpStream;

#[tokio::main]
async fn main() {
  let addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), 6142);
  let mut stream = TcpStream::connect(addr).await.unwrap();

  let _result = stream.write(b"hello world\n").await;
}
```

content/docs/getting-started/hello-features.md
ultrasaurus: When first using tokio, I found it really difficult to construct code using `AsyncReadExt` and `AsyncWriteExt` (though maybe better docs would have helped!).  I do feel a little unfortable with the *magic* of prelude, but it does make the examples easier to write.

@LucioFranco can you say more about why you dislike preludes?  and why explicit `use` is a good thing and not just more typing?

content/docs/getting-started/hello-features.md
jenmei: Yeah, the removal of the prelude wasn't necessary but I thought was similarly themed – we are using only what we need instead of the other things the prelude gives you. I also agree with @LucioFranco (I think) in that I think preludes are an unnecessary extra level of indirection. I can see them being useful if it's saving you a lot of boilerplate, but in this case it's just exchanging one line for another.

content/docs/getting-started/hello-features.md
jenmei: Not sure what you mean? The hello world example has `mut`. If I take it out, I get:

```
[hello-world (master)]$ cargo run 
   Compiling hello-world v0.1.0 (/Users/jenmei/src/rust/tmp/hello-world)
error[E0596]: cannot borrow `stream` as mutable, as it is not declared as mutable
 --> src/main.rs:9:16
  |
6 |   let stream = TcpStream::connect("127.0.0.1:6142").await.unwrap();
  |       ------ help: consider changing this to be mutable: `mut stream`
```

content/docs/getting-started/hello-features.md
jenmei: Sorry, what I wrote above doesn't make sense even to me.

What I meant to say is replace this:

```md
If we change our code, so that it only uses the `#[tokio::main]` macro:
```

with 
```md
If we replace our `src/main.rs` with this (basically hello world w/ with the `#[tokio:main]` macro:
```

Maybe it's clear to others but the way it was written I wasn't sure whether you meant to replace the whole file or just the main function.

content/docs/getting-started/hello-features.md
ultrasaurus: sorry about that, you are correct!

content/docs/getting-started/hello-features.md
ultrasaurus: added some narrative to address this.  I referenced some Rust doc I found that says that prelude in crates allows people to easily include whatever is always needed (and I don't think anyone would ever write such a useless app).  In any case, I think it is at least explained.

content/docs/getting-started/hello-features.md
ultrasaurus: I added some text to provide a bit of a better explanation here.  As long as tokio has a prelude feature, I think hello-world example should use it, but agree that when we're talking about using *specific* features, people probably aren't looking to use prelude!

content/docs/getting-started/hello-features.md
ultrasaurus: fixed

ultrasaurus:
I think I've addressed all the feedback.  I would like someone to proof-read the text and plan to write issues for the TODOs in the PR description above.

content/docs/getting-started/hello-world.md
```diff
@@ -1,150 +1,154 @@
...
-[`write_all`]: https://docs.rs/tokio/0.2/tokio/io/trait.AsyncWriteExt.html#method.write_all
-[`AsyncWriteExt`]: https://docs.rs/tokio/0.2/tokio/io/trait.AsyncWriteExt.html
-[full-code]: https://github.com/tokio-rs/tokio/blob/master/examples/hello_world.rs
-[Netcat]: http://netcat.sourceforge.net/
-[Nmap.org]: https://nmap.org
+Now that we have successfully built a tiny client to get a feel for network
+programming with Tokio, we'll dive into more detail about how everything works.
+
+[Rust programming language]: https://www.rust-lang.org/
+[Tokio 0.1 docs]: https://v0-1--tokio.netlify.com/
```
jenmei: Is this the url we want to use? I was thinking it might go to `https://docs.rs/tokio/0.1.22/tokio/index.html#modules`?

content/docs/getting-started/hello-world.md
```diff
@@ -1,150 +1,154 @@
...
+As our first introduction to [asynchronous programming](../overview) with
+Tokio, we will start with a tiny app that sends "hello, world!" over a
+network connection.
 
-```toml
-[dependencies]
-tokio = { version = "0.2", features = ["full"] }
-```
+The difference between Tokio and a Rust program written using the standard
+library is that Tokio has the flexibility to block program execution
```
jenmei: This doesn't read right to me. It sounds like we're saying that Tokio's advantage is that it blocks your program. Which a) doesn't sound like a good thing; and b) not sure that's the case. It only blocks execution within the async function, not the whole program, right?

content/docs/getting-started/hello-world.md
```diff
@@ -1,150 +1,154 @@
...
-async fn main() {
-    // application comes here
-}
+Install [`socat`](../network-utilities/socat), which is a network utility that
+we'll use to simulate a server. Then type the following command to print
+out everything that is received on port 6142 (a somewhat arbitrary number
+we have chosen for this example):
+
+```bash
+socat TCP-LISTEN:6142 stdout
```
jenmei: Completely unnecessary, but I would suggest changing the command to `socat TCP-LISTEN:6142,fork stdout` so that they don't need to restart `socat` each time their program runs.

content/docs/getting-started/hello-world.md
```diff
@@ -1,150 +1,154 @@
...
 ```
 
-Here we use Tokio's own [`io`] and [`net`] modules. These modules provide the same
-abstractions over networking and I/O-operations as the corresponding modules in
-`std` with a difference: all actions are performed asynchronously.
+An easy way to simulate a client (for a text-based protocol) is to use
+[`telnet`]. To connect to our simulated server, we'll open a different
```
jenmei: Sorry for not providing this feedback before, but I would suggest using netcat instead of telnet. I think `nc` is now more likely to be found on OSes (including Mac OS X) without a separate install. Also, as network code debugging tool, telnet can have some weird side effects (because it's primarily written for the telnet protocol, which is almost, but not quite, plain text; see the [Wikipedia page](https://en.wikipedia.org/wiki/Telnet#Telnet_data)).


for network programming, telnet can have some weird.

[telnet is weird][https://superuser.com/questions/1461609/what-is-the-difference-between-telnet-and-netcat].

content/docs/getting-started/hello-world.md
```diff
@@ -1,150 +1,154 @@
...
-# use tokio::net::TcpStream;
-#[tokio::main]
-async fn main() {
-    // Connect to port 6142 on localhost
-    let stream = TcpStream::connect("127.0.0.1:6142").await.unwrap();
+Let's quit telnet and write some Rust code to send some text to our server.
+To *escape* from our TCP session (`^]`), we need to hold down `Ctrl` key and
+type `]`). Then at the telnet prompt (`telnet >`), typing `quit` will close
+the connection and exit the program.  (Closing the Telnet/TCP connection
+also causes socat to quit.)
```
jenmei: With `nc` exiting is just `ctrl-d` so less to read. :-)

content/docs/network-utilities/telnet.md
```diff
@@ -0,0 +1,41 @@
...
+Telnet utility allows users to test connectivity to remote machines and
+issue commands through the use of a keyboard, and one of the simplest ways to
+check if a port is open and listening on a network.
+
+# How to install
+
+## Debian / Ubuntu Linux
+
+```bash
+sudo apt-get update && sudo apt-get install telnet
```
jenmei: Pretty sure `nc` is included on Debian and Ubuntu by default, but haven't confirmed.

content/docs/network-utilities/telnet.md
```diff
@@ -0,0 +1,41 @@
...
+## Fedora, CentOS, RedHat
+
+```bash
+sudo yum install telnet
+```
+
+## Mac Homebrew
+
+```bash
+brew install telnet
```
jenmei: `nc` is included on Mac OS X (`telnet` has been removed starting with High Sierra).

content/docs/getting-started/hello-world.md
ultrasaurus: I think a pointer to the site will be helpful, then people have the guides too!

content/docs/getting-started/hello-world.md
ultrasaurus: oops -- missed "not" changing to "has the flexibility to avoid blocking program execution"

content/docs/getting-started/hello-world.md
ultrasaurus: I like this!

ultrasaurus:
added consideration for `nc` to issue about testing network utils: https://github.com/tokio-rs/website/issues/407

PR #404, opened 2020-01-14
fix links, remove duplicate echo example

ultrasaurus:
if this is controversial, I can put into two PRs...

1) API doc links were pointing to specific version of Tokio, now they point to latest
2) removed echo example here -- it's hard to keep this in sync with [README](https://github.com/tokio-rs/tokio#example) + [echo server tutorial](https://tokio.rs/docs/getting-started/echo/).  I figure one fewer copy of this would make things easier to maintain (and not sure it adds a ton of value here)

Note; there are some key content sections which are out of date, and I've opened those up as a separate issue https://github.com/tokio-rs/website/issues/405, since that might take a bit more time for someone to write

### DIFF
--- a/content/docs/overview.md
+++ b/content/docs/overview.md
@@ -14,17 +14,27 @@ More specifically, Tokio is an event-driven, non-blocking I/O platform
 for writing asynchronous applications with Rust. At a high level, it
 provides a few major components:
 
-* A multithreaded, work-stealing based task [scheduler].
-* A [reactor] backed by the operating system's event queue (epoll, kqueue,
-  IOCP, etc...).
-* Asynchronous [TCP and UDP][net] sockets.
+* Tools for [working with asynchronous tasks][tasks], including
+  [synchronization primitives and channels][sync] and [timeouts, delays, and
+  intervals][time].
+* APIs for [performing asynchronous I/O][io], including [TCP and UDP][net] sockets,
+  [filesystem][fs] operations, and [process] and [signal] management.
+* A [runtime] for executing asynchronous code, including a task scheduler,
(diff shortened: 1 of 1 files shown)


content/docs/overview.md
```diff
@@ -147,51 +147,8 @@ tasks.  Tokio implements [async] syntax to improve readability of futures.
...
-    // Start the Tokio runtime
-    tokio::run(server);
-}
-```
-
-More examples can be found [here](https://github.com/tokio-rs/tokio/tree/v0.1.x/tokio/examples).
+# Get started with Tokio
+
+If you like to read code first, complete examples can be found
+[here](https://github.com/tokio-rs/tokio/tree/*/tokio/examples), or
```
jebrosen: This link 404s for me (docs.rs links above works because docs.rs has special support for it)

ultrasaurus: Oops -- I should have tested each link.  I'll fix.  Thank you!

ultrasaurus: fixed!

ultrasaurus:
Also, https://docs.rs/tokio/0.1/tokio/reactor/index.html no longer exists (https://docs.rs/tokio/*/tokio/reactor/index.html).  What should this be instead?  

> A reactor backed by the operating system’s event queue (epoll, kqueue, IOCP, etc…).

jebrosen:
That text looks like a near-copy of https://docs.rs/tokio/0.1.22/tokio/, the latest version of which (https://docs.rs/tokio/0.2.9/tokio/) doesn't refer to reactors at all. "I/O driver based on epoll/kqueue/etc" looks like the best match to what used to be called the (I/O) reactor.

Issue #409, opened 2020-01-14
Links to guides is broken in the readme and docs


bdurrani:
Readme points to `https://tokio.rs/docs/` when you click on the guides

As per my convo with @ultrasaurus in discourse

```
for now, you can navigate from: https://tokio.rs/
but I think it would be friendly to redirect https://tokio.rs/docs to the start page, which is currently https://tokio.rs/
```

PR #414, opened 2020-01-24
netlify redirect for /docs => docHome

ultrasaurus:
I didn't find a good answer to this in hugo docs, so trying out netlify config to redirect

to fix https://github.com/tokio-rs/website/issues/409


ultrasaurus:
this didn't work at all... will have to look more at Netlify docs to figure out why

PR #430, opened 2020-07-21
Adds SVG favicon

sl4m:
This adds an SVG favicon using the existing Tokio logo here: https://github.com/tokio-rs/website/blob/master/public/img/icons/tokio.svg. It's resized and supports dark mode (using CSS media query). Website falls back on the PNG favicon if the browser doesn't support SVG.

Tested against Chrome, Firefox, and Safari on macOS.

Screenshot 2020-07-21 at 10 44 00Screenshot 2020-07-21 at 10 43 44

Currently, the PNG favicon does not support dark mode. Ideally, this favicon should be modified to work on either light or dark mode. Something to consider in the future.

### DIFF
--- a/components/layout.tsx
+++ b/components/layout.tsx
@@ -13,6 +13,8 @@ const Layout: FC<Props> = ({ blog, children }) => (
     <Head>
       <title>Tokio</title>
       <meta name="viewport" content="width=device-width, initial-scale=1" />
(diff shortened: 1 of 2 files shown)


sl4m:
> The dark-mode support is neat! Thank you! I tested it against the newest IE and Edge on Windows 10, and firefox/chrome on Linux.

Thank you for reviewing and testing it on the other platforms!

PR #541, opened 2021-01-20
Import futures::task::self in async chapter

seanchen1991:
Hello, I noticed that the mini tokio tutorial code didn't work unless I added a `use futures::task::self;` statement. It looks like `task::self` is included in the code snippet, but it's commented out and doesn't render on the site itself. 

This PR adds the necessary import statement in one possible place, namely the `impl ArcWake for Task` block. The statement could alternatively be added in the `impl MiniTokio` block a bit further down.


Darksonn (approved):
Thanks

PR #546, opened 2021-02-06
Fix link URL and typo in tutorial

magurotuna:
Currently the link to the `copy` version of echo server seems wrong. If people click this link, they are brought to the `manual` example. So in this PR I've fixed it.
In addition, also fixed a typo :)


Darksonn (approved):
Thanks!

PR #547, opened 2021-02-07
Specify that two terminal windows are required

nahuakang:
I'm going through tokio's tutorial and thought it'd be better to specify that the reader must open a new terminal window to try the `$ mini-redis-cli get foo` command. Let me know what you guys think :)


Darksonn (approved):
Thanks!

PR #551, opened 2021-02-18
tutorial: Fix crate name for mini-redis

starthal:
The `cargo install` command did not refer to the actual crate name.


Darksonn (approved):
Thanks

PR #552, opened 2021-02-20
Upgrade crossbeam to 0.8.0

magurotuna:
I noticed the version of crossbeam to be outdated. This just upgrades crossbeam to 0.8.0.


taiki-e (approved):
Thanks

PR #555, opened 2021-02-22
Fix select chapter in tutorial

magurotuna:
Reading the "select" chapter in the tutorial, I was puzzled a bit by the description that says `Future` is a trait that must be implemented by a value in order to call `.await` on it, because the previous chapter "Async in depth" already covers that fact. So I attempted to fix it.
Besides, fixed a tiny typo as well.


Darksonn (approved):
Thanks!

PR #568, opened 2021-05-01
Remove unneeded mutability on TcpListener

bertonha:
Going to the tutorial I noticed this warning

```
warning: variable does not need to be mutable
 --> src\bin\echo-server.rs:6:9
  |
6 |     let mut listener = TcpListener::bind("127.0.0.1:6142").await.unwrap();
  |         ----^^^^^^^^
  |         |
  |         help: remove this `mut`
  |
  = note: `#[warn(unused_mut)]` on by default
```


Darksonn (approved):
Thanks

PR #582, opened 2021-06-29
Clarify which values the MutexGuard refers to

joshchoo:
In the "**Holding a MutexGuard across an .await**" section of the [Shared state tutorial](https://tokio.rs/tokio/tutorial/shared-state), it took me a while to figure out that `MutexGuard` referred to the `lock` (it wasn't stated explicitly in the tutorial). I understood only after I ran the code sample. We can make it clearer for readers by specifying the `MutexGuard` type on `lock`.


Darksonn (approved):
This seems fine to me.

PR #584, opened 2021-07-03
Show fewer blogs in sidebar and link to an index

cab:
## Changes

This adds a new blog index at `/blog`. It lists all blog posts for all years.

A new link is added to the blog post page sidebar. Only the 6 most recent posts are shown in the sidebar.

I added a new `blog.js` module to encapsulate this logic, and pulled out most of the related code in `[...slug].js`.

I also added a fix for the "fs module not found" error that was popping up without related changes. This might be specific to my environment, and I can take that out if it's not wanted or too unrelated.

### Blog Post Sidebar
image

### Blog Index
image


Darksonn (approved):
Thanks for doing this!

PR #585, opened 2021-07-03
fix blog index and link on mobile

cab:
@davidpdrsn caught that #584 looked incorrect on mobile. My bad, I should have tested that out.

This fixes the styling on mobile by making the index full-width and only showing the "All Posts" link when the TOC is expanded.

### Blog Post
image
image

### Index
image


Darksonn (approved):
Thanks.

PR #597, opened 2021-08-02
To not -> Not to

boehs:
Not to is a split infinitive and is historically frowned upon. This simple change improves readability greatly (source: it's 12 in the morning)


Darksonn (approved):
Thanks.

PR #609, opened 2021-09-10
Center `Getting Started` button on landing page

### LINKED ISSUE #437: Center "get started" button on mobile
It is currently aligned to the left.

nylonicious:
Closes #437.

Alternative solution is to remove `has-text-left-mobile` instead of change in the pr. 
https://github.com/tokio-rs/website/blob/1d940c88f160c2fe0d657a6e43e10a406eb25a89/components/hero.tsx#L25

How it looks with only button centered.
just-button

How it looks with entire container centered on mobile.
everything

I've found everything centered a bit harder to read so I didn't include it. But if entire container centered looks better I'll change it.

### DIFF
--- a/components/hero.tsx
+++ b/components/hero.tsx
@@ -37,9 +37,11 @@ const Hero: FC = () => (
                   network applications. It leverages Rust's ownership and
                   concurrency model to ensure thread safety. */}
           </h2>
-          <a href={gettingStarted} className="button is-link is-medium">
-            Get Started
-          </a>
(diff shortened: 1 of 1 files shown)


hawkw:
Hmm, I agree that the centered text feels a harder to read. I have a slight preference for leaving it the way you have it now. But, it would be nice to hear from others on that.

Darksonn (approved):
Ok, let's go with what we have here then.

PR #623, opened 2021-11-21
chore: fix a typo

samuelxing-okta:
To inline with https://github.com/tokio-rs/website/blob/master/tutorial-code/channels/src/main.rs#L66


Darksonn (approved):
Thanks.

PR #647, opened 2022-04-01
Update io.md, minor clarification

createyourpersonalaccount:
Clarifies that the next code block relates to the server and not client that was
just mentioned previously.


Darksonn (approved):
Seems good to me.

PR #675, opened 2022-10-11
Add Cloudflare logo

divyeshio:
Added cloudflare logo to the Logos section on the landing page, as Cloudflare uses tokio for Pingora, which can be cited [here](https://blog.cloudflare.com/how-we-built-pingora-the-proxy-that-connects-cloudflare-to-the-internet/#:~:text=The%20Tokio%20async%20runtime%20turned%20out%20to%20be%20a%20great%20fit%20for%20our%20needs)


carllerche:
Thanks for the PR. To add logos, we need to get explicit approval from a representative of the organization with authority to approve the logo usage. The set of logos on the homepage is not intended to be exhaustive, more of a sample of who is using it.

PR #676, opened 2022-10-27
blog: announce async-backtrace

jswrenn:
Announcement post for initial release of async-backtrace crate.

@carllerche

### DIFF
--- a/content/blog/2022-10-announcing-async-backtrace.md
+++ b/content/blog/2022-10-announcing-async-backtrace.md
@@ -14,8 +14,19 @@ In synchronous, multi-threaded applications, you can investigate deadlocks by
 inspecting stack traces of all running threads. Unfortunately, this approach
 breaks down for most asynchronous Rust applications, since suspended tasks —
 tasks that are not actively being polled — are invisible to traditional stack
-traces. The async-backtrace crate fills this gap, allowing you see the state of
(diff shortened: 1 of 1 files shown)


carllerche:
Rendered: https://deploy-preview-676--tokio.netlify.app/blog/2022-10-announcing-async-backtrace

carllerche:
I think we should expand on how this crate relates to other projects. Readers will be asking themselves two questions: "How does this compare to X (`tracing`, "Tokio Console", ...)  other lib" and "How does this fit into my workflow".

The answer to the first question is that it is complementary and is a "building block" utility (better wording needed). Eventually, Tokio Console could use this.

LucioFranco:
The example feels kinda hard to follow maybe include more of the example and some more complex real world examples? Like can you show this with hyper?

jswrenn:
I've updated the post to include a totally self-contained example, and updated the repo to contain many more examples.

Since the usage guidelines of the crate are fairly simple — apply `#[async_backtrace::framed]` as liberally as possible —  I don't think the cognitive load of introducing additional libraries would be justified by its demonstrative benefits. The usage would be the same, but aspects of the example would be unfamiliar to anyone without experience with hyper.

carllerche (approved):

PR #678, opened 2022-11-16
Fix query key in channels tutorial

palango:
This brings the tutorial code to the state of the linked code.

It might be nicer to use `hello` as the key and `world` as the value as in the earlier tutorials though.


Darksonn (approved):
Thanks.

PR #690, opened 2023-01-24
fix(blog): update default instrument level to `info`

leungyauming:
> Unless overriden, a span with the [INFO](https://docs.rs/tracing/latest/tracing/struct.Level.html#associatedconstant.INFO) [level](https://docs.rs/tracing/latest/tracing/struct.Level.html) will be generated. The generated span’s name will be the name of the function. By default, all arguments to the function are included as fields on the span. Arguments that are tracing [primitive types](https://docs.rs/tracing/latest/tracing/field/trait.Value.html#foreign-impls) implementing the [Value trait](https://docs.rs/tracing/latest/tracing/field/trait.Value.html.) will be recorded as fields of that type. Types which do not implement Value will be recorded using [std::fmt::Debug](https://doc.rust-lang.org/nightly/core/fmt/trait.Debug.html).

See: [docs on `tracing::instrument`](https://docs.rs/tracing/latest/tracing/attr.instrument.html#)

### DIFF
--- a/content/tokio/topics/tracing.md
+++ b/content/tokio/topics/tracing.md
@@ -152,7 +152,7 @@ fn trace_me(a: u32, b: u32) -> u32 {
 
 Each invocation of `trace_me` will emit a `tracing` Span that:
 
-1. has a verbosity [level] of `trace` (the greatest verbosity),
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
Thanks.

PR #692, opened 2023-02-03
Use a raw string literal when including markdown files during doc test creation

PigeonF:
When trying to run the tests on Windows, the compilation fails, because the paths to the markdown files get canonicalized to contain backslashes (e.g. `\\?\D:\tokio-rs\website\content\tokio\index.md`). This leads to an `unknown character escape` error. Using a raw string literal fixes this by having rustc interpret the path as-is.


Darksonn (approved):
Thanks.

PR #693, opened 2023-02-14
added instructions to create a new file connection.rs

anupj:
It was not clear (to me) when I was following the tutorial that I was supposed to create a new `connection.rs` file to add the code for the connection struct. Hopefully this will make it clearer.


Darksonn (approved):
Thanks.

PR #697, opened 2023-04-14
Fix side navigation menu overflowing on narrow screens

karen-sarkisyan:
Found this a little annoying that on my 13'' MacBook vertical menu overflows and it's impossible to scroll it until the end of the page.

before

So I made the container sticky (instead of nested `aside` element) and overflow-able with scroll. Also rearranged paddings so that they are on container level too, I think it's slightly better for how the page feels.

after

### DIFF
--- a/styles/styles.scss
+++ b/styles/styles.scss
@@ -520,7 +520,11 @@ a:active {
     }
 
     @include tablet {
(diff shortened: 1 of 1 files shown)


Darksonn (approved):
Thanks.

PR #702, opened 2023-05-02
Dropping repetitive text

primeagen-rustaceans:
The Line 1 and Line 2 conveys the same idea of using full feature flag for this tutorial. Line 1. When depending on Tokio for this tutorial, the `full` feature flag is enabled. Line 2. For now, use the "full" feature when depending on Tokio.


Darksonn (approved):
Thanks.

PR #703, opened 2023-05-02
Update in spawning.md

primeagen-rustaceans:
Updating document tone for a better understanding. Since an info column is separating in between the reading tone of the two paragraphs, keeping the TcpListener information in the former and other information in the latter one.


Darksonn (approved):
Thanks.

PR #705, opened 2023-05-23
Update index.md

primeagen-rustaceans:
The context "Tokio Topics" is self explanatory. By removing it making the language of text more formal.


Darksonn (approved):
I don't understand how this makes it _more formal_, but I agree that it is an improvement nonetheless.

PR #718, opened 2023-10-03
Clarify variable naming in async.md

skaunov:
It's a little bit confusing for a newcomer that notify one method is called on notify two -- it gives wrong hint/intuition to some readers, so it worth to choose a name which won't suggest such semantic, which isn't there.

Looks like this part of code is only in tutorial text, and nothing to sync in the supplementary code.


Darksonn (approved):
Thanks.

PR #722, opened 2023-10-11
Tutorial: Add link to rayon example

matildasmeds:
There are great resources around, and the tutorial could be the first contact point for many people with Tokio. 
Why not link them? I think practical examples increase the overall value of the tutorial.


Darksonn (approved):
Thanks.

PR #725, opened 2023-10-23
Update links in footer

tmsagarofficial:
Ensure all the links in the footer open in a new tab, improving the user experience.
As otherwise the user would have to come back to this tab everytime they click on a link to be able to navigate further.


Darksonn (approved):
Thanks.

PR #730, opened 2023-10-31
Add syntax highlight for sample codes in glossary section

xring:
add syntax highlight for sample codes in glossary section


Darksonn (approved):
Thanks.

PR #741, opened 2023-12-25
Update hello-tokio.md to use Client

erikschul:
`client` doesn't exist, but `Client` does.

### DIFF
--- a/content/tokio/tutorial/hello-tokio.md
+++ b/content/tokio/tutorial/hello-tokio.md
@@ -31,13 +31,13 @@ mini-redis = "0.4"
 Then, open `main.rs` and replace the contents of the file with:
 
 ```rust
-use mini_redis::{client, Result};
+use mini_redis::{Client, Result};
 
 # fn dox() {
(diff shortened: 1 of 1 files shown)


hds (changes requested):
Hi @erikschul, thanks for your PR!

The code here is actually correct for the version of `mini-redis` being used. In the `Cargo.toml` section a little further up, we specify `mini-redis = "0.4"`, this will resolve to version 4.1, which is the latest one on crates.io.

However, as you have seen, the latest version on crates.io is actually pretty old and isn't the same as the `master` branch. You can see in the `lib.rs` from the commit that was released as 4.1 that there is still a module called `client` there:

https://github.com/tokio-rs/mini-redis/blob/b3926d40e38dd0e4880c5ddf5880b5fe4f908cdc/src/lib.rs#L29

Since we have breaking API changes on `master`, the next release of `mini-redis` would have to be 0.5 and then this example could be updated explicitly (it wouldn't automatically pick up a 0.5 version).

PR #742, opened 2023-12-27
fix(shared-state): use std::sync::Mutex which not implemented Send

swcumt:


hds (changes requested):
Thanks for your PR!

The `Mutex` from tokio isn't a drop in replacement for the one from the std library. It needs to be awaited (and doesn't return a `Result`). That means that this change won't compile.

I'm not quite sure what the objective of this PR is trying to achieve though. Would you be able to add a description of what the change is doing to the PR and then we can work together to find a solution.

The code here is taken from the package in the `tutorial-code` directory:
https://github.com/tokio-rs/website/blob/d49064b5c94754fffa4449f9138d8d5f7b65d79d/tutorial-code/shared-state/src/main.rs

So you can use that space to test out any changes to ensure that they compile and run correctly. The code there should also be updated to match any changes in the markdown files.

PR #751, opened 2024-02-26
fix: `# if true { return }` is not hidden in html code block

PeterD1524:
`# if true { return }` is shown in the code block of the topic page [Unit Testing](https://tokio.rs/tokio/topics/testing):

Based on https://github.com/tokio-rs/website/pull/735, `# if true { return }` was used to let doc tests pass.
The original function to remove `#` lines does not cover all cases.
This PR fixes that function to match [what librustdoc does](https://github.com/rust-lang/rust/blob/0ecbd0605770f45c9151715e66ba2b3cae367fcb/src/librustdoc/html/markdown.rs#L170).

Another forgotten `#` on the same page:

Also in blog [New Timer implementation](https://tokio.rs/blog/2018-03-timers):

These `#` lines will be removed by the fixed function.

### DIFF
--- a/content/tokio/topics/testing.md
+++ b/content/tokio/topics/testing.md
@@ -166,7 +166,7 @@ be used as a mock:
 #         line.clear();
 #     }
 # }
-
+#
 #[tokio::test]
 async fn client_handler_replies_politely() {
     let reader = tokio_test::io::Builder::new()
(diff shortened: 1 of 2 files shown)


Darksonn (approved):
Thanks.

PR #759, opened 2024-04-06
Add GitHub actions step for validating prettier code style

junlarsen:
This patch adds a small GitHub actions job for verifying that the code style conforms with the prettier configuration.

I have also ran prettier locally to fix up the files that didn't conform with the code style prior to this patch.

Let me know what you think. I know this isn't all that important, but it bothered me that code unrelated to my changes would re-format in my editor.


Darksonn (approved):
LGTM.

PR #762, opened 2024-04-22
Fixed watch

takassh:
`Clone` of `watch::Sender` was introduced in https://github.com/tokio-rs/tokio/pull/6388, and that change was shipped in the [latest v1.37.0 release](https://github.com/tokio-rs/tokio/releases/tag/tokio-1.37.0).
So I fixed the document


mox692 (approved):
Thanks!

PR #763, opened 2024-05-01
Queing spelling change in channels.md

garikAsplund:
Both 'queing' and 'queueing' are correct, but for consistency's sake should stick to 'queing'.


Darksonn (approved):
Thanks.

PR #765, opened 2024-05-20
Remove link to dashmap

matildasmeds:
Dashmap seems to be a bit risky to use with tokio. Yet, according to observations on Tokio Discord: people keep trying, and maintainers keep recommending against. I think ideally we would not link to this crate from tokio docs.


matildasmeds:
Closing this one, as per discussions on Tokio discord. Will file another PR with adequate warnings about Send guards later.

PR #767, opened 2024-06-03
Fix grammar 2023-11-27-announcing-axum-0-7-0.md

Gluroda:
The subject "high level utilities" is plural, so the verb should be "have" instead of "has."


Darksonn (approved):
Thanks.

PR #769, opened 2024-07-11
Update link of trait BufMut in framing.md

bangjiehan:
https://docs.rs/bytes/1.0.0/bytes/trait.BufMut.html worked and https://docs.rs/bytes/1.6.0/bytes/trait.BufMut.html failed but i don't know why.


Darksonn (approved):
Thanks.

PR #770, opened 2024-07-16
Corrected overflow-x style to remove unnecessary horizontal scroll bar

terrence-ou:
Hi,

When I was browsing the website, I found unnecessary horizontal scroll bars under each of the code blocks, which kind of affect visual consistency, as the following image shows:

x-bar

Then I went to the `styles.scss` file and found the attribute `overflow-x` was set to `scroll`. This means the horizontal scroll bar will always be there, whether the content is overflowing on the x-axis or not. I changed the value to `auto` to hide the bar when no overflow happens but show the scroll bar when horizontal scrolling is necessary. 
The result after this modification is as shown in the following images:

When no x-scrolling needed:
no-x-bar

When content overflows:
Screenshot 2024-07-15 at 10 17 46 PM

Thanks so much for considering this update!

### DIFF
--- a/styles/styles.scss
+++ b/styles/styles.scss
@@ -934,7 +934,7 @@ Docco style used in http://jashkenas.github.com/docco/ converted by Simon Madine
 
 .hljs {
   display: block;
-  overflow-x: scroll;
+  overflow-x: auto;
 }
 
 .hljs-comment,


Darksonn (approved):
This seems to be browser-dependent. I don't get the scroll bars without the change.

Either way, thank you.

PR #780, opened 2024-10-18
use node.js 20.18.0

joshka:
This is the current LTS version which is recommended for prod usage.

Updates the netlify config and adds a mise.toml config file to configure
the node.js version to use when running the site locally.

This will fix the problem running next.js in https://github.com/tokio-rs/website/pull/778

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -1,0 +1,2 @@
+[tools]
+node = "lts"
--- a/.mise.toml
+++ b/.mise.toml
@@ -1,2 +1,0 @@
-[tools]
-node = "lts"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,7 +52,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: "18"
+          node-version: "20"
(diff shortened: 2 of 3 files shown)


.mise.toml
```diff
@@ -0,0 +1,2 @@
+[tools]
+node = "lts"
```
mox692, line 2: Is it possible to make this change as a separate issue or PR?

mox692 (commented):
Thanks!
Regarding the version bump, I confirmed locally that there is no issue.

.mise.toml
joshka: Sure. I've removed it and probably wouldn't bother with a PR to add it.

mox692 (approved):
Can we also update the ci settings? Otherwise lgtm.

https://github.com/tokio-rs/website/blob/3c1522cd856217808d8414e88bf389a4d9589357/.github/workflows/ci.yml#L55

joshka:
> Can we also update the ci settings? Otherwise lgtm.
> 
> https://github.com/tokio-rs/website/blob/3c1522cd856217808d8414e88bf389a4d9589357/.github/workflows/ci.yml#L55

done

PR #792, opened 2024-10-28
Update Next.js to v15

joshka:
- **bump caniuse db**
- **fix sed command to work with both GNU and MacOS sed**
- **Update to next.js 14**
- **Upgrade to nextjs 15**


Darksonn (approved):
Thank you.

Issue #461, opened 2020-07-25
In hello-tokio, add note to check for .await if code seems to not run 


carllerche:
A common beginner mistake is to forget to call .await. The symptom is that the program does not do the expected work.

We should call this out on the hello-tokio page in a “troubleshooting” section maybe. We should also call out the must_use lint that should catch this error.

Issue #401, opened 2020-01-13
[topic] cargo dependencies: `features = ...`


ultrasaurus:
It would be great to have a page that explains each feature group, ideally with an example for each feature group.

PR #801, opened 2024-11-26
Simplify hello-tokio, move intro article

joshka:
I moved the intro to tokio introdcution that was at /tokio/tutorial into its own section as it contains no tutorial specific info.

The tutorial/setup page is now at /tutorial (and a redirect is added from the original page)

A bunch of content out of hello-tokio has been extracted to topics/{async,feature-flags}

Fixes: https://github.com/tokio-rs/website/issues/461
Related: https://github.com/tokio-rs/website/issues/401
Related: https://github.com/tokio-rs/website/issues/503#issuecomment-2501987552


joshka:
TOC:

image

content/tokio/topics/feature-flags.md
Darksonn, line 1: This is a pretty small page. Perhaps we could just link directly to https://docs.rs/tokio/latest/tokio/#feature-flags without this intermediate page?

content/tokio/topics/async.md
Darksonn, line 1: Can you elaborate on what happened here?

Generally, articles that under `topics/` end up being stuff that people will read *after* reading the tutorial, but this seems like material they should see at the beginning of the tutorial.

content/tokio/topics/feature-flags.md
joshka: I'm ok with doing that. There was a mention of adding this page in https://github.com/tokio-rs/website/issues/401, so I figured it would be worth keeping.

content/tokio/topics/async.md
joshka: I pulled this out of the tutorial intentionally. IMO, It feels like the hello-tokio page goes too deep too soon, and shows a bunch of code which is demonstrative rather than instructive. It's trying to serve two purposes and in doing so it diminishes each of them. For readers that are looking for just the information, finding it in the context of a tutorial is weird, and for readers doing a tutorial to learn about how to write tokio apps it's a bit much info that's not necessary to achieve the goal. I think the info should stick around though, so I think it should be a topic. I tried to make it not seem like part of the tutorial (even though it references it a bit still).

Generally, I think it would be worthwhile thinking about the topics/ pages as being things which might simplify answers in discussions (saving keystrokes and increasing the reach of the docs). Rather than a full explanation, a smaller "It's because blah, see /topics/blah" type response might be appropriate. I think I'd like to see some small recipe type things (I've noticed a bunch of these just by reading the discord channels).

content/tokio/tutorial/hello-tokio.md
```diff
@@ -13,19 +13,36 @@ then read back the key. This will be done using the Mini-Redis client library.
...
 
 ```toml
-tokio = { version = "1", features = ["full"] }
-mini-redis = "0.4"
+[dependencies]
+tokio = { version = "1.41.1", features = ["full"] }
+mini-redis = "0.4.1"
 ```
 
+We use the `full` feature flag in the example for simplicity. For more
```
Darksonn, line 42:
```suggestion
We use the `full` feature flag for simplicity. For more
```

content/tokio/topics/async.md
```diff
@@ -0,0 +1,134 @@
...
+gets transformed into:
+
+```rust
+fn main() {
+    let mut rt = tokio::runtime::Runtime::new().unwrap();
+    rt.block_on(async {
+        println!("hello");
+    })
+}
+```
```
Darksonn, line 134: It would be nice to have a link to the documentation of the `#[tokio::main]` macro. The link could say something along the lines of "to see all of the options you can specify when using `#[tokio::main]`, see the documentation" or similar.

content/tokio/tutorial/hello-tokio.md
```diff
@@ -94,155 +157,58 @@ The [`client::connect`] function is provided by the `mini-redis` crate. It
...
-fn main() {
-    let mut rt = tokio::runtime::Runtime::new().unwrap();
-    rt.block_on(async {
-        println!("hello");
-    })
-}
+```rust,ignore
+# let result = 42;
+println!("got value from the server; result={result:?}");
 ```
```
Darksonn, line 203: This `let result = 42` line shows up in the rendered html for me. Perhaps you need to drop the `ignore` from the codeblock to fix this?

PR #827, opened 2025-06-20
Use UTC functions on date objects

aditsachde:
Currently, the sidebar on blog pages shows the wrong date for the posts – they are all off by one.

The markdown files have pages with no specified timezone, so `new Date(post.date)` parses these in UTC. However, the `getDate/Month/Year` functions use the local timezone. This makes the dates that show up on the sidebar dependent on the current time of the build machine.

Fixed:

image

Previously: 

image


Darksonn (approved):
Thanks.

PR #860, opened 2026-06-23
docs(tutorial/io): clarify why shutdown must be called after io::split

francesco-gaglione:
`io::split` wraps the writer in Arc<Mutex<...>> and does not close the connection on drop, so EOF must be signaled explicitly via `shutdown()`.


Darksonn (approved):
Thanks.

